PR #1, opened 2024-05-18
chore: ts-docs for ccc core function

SpectreMercury:
- [x] address
- [x] bytes
- [x] ckb
- [x] client
- [x] fixedPoint
- [x] hasher
- [x] hex
- [x] num
- [x] signer
- [x] utils

### DIFF
--- a/packages/core/src/address/address.advanced.ts
+++ b/packages/core/src/address/address.advanced.ts
@@ -14,7 +14,7 @@ import { type AddressLike } from "./index";
  *
  * @example
  * ```typescript
- * const addressInfo = addressPayloadFromString("bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kygt080");
+ * const addressInfo = addressPayloadFromString("ckt1112139193129");
  * console.log(addressInfo.prefix); // Outputs the address prefix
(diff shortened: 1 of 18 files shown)


packages/core/src/address/address.advanced.ts
```diff
@@ -4,6 +4,23 @@ import { Client, KnownScript } from "../client";
...
+ * Parses an address string into an address information object.
+ *
+ * @param address - The address string to parse.
+ * @returns An object containing the address prefix, address format, and payload array.
+ *
+ * @throws Will throw an error if the address format is unknown.
+ *
+ * @example
+ * ```typescript
+ * const addressInfo = addressPayloadFromString("bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kygt080");
```
Hanssen0, line 17: This example doesn't seem to be correct.

Hanssen0 (changes requested):

packages/core/src/address/address.advanced.ts
SpectreMercury: yep, seem like we should use ckb address here

PR #3, opened 2024-06-17
feat: if wallet not installed, redirect to download page

SpectreMercury:
if wallet is not installed redirect to wallet download page


Hanssen0 (approved):
lgtm

PR #6, opened 2024-06-25
fix: demo lumos config environments

SpectreMercury:
- change demo lumos config change depeneds on signer client


Hanssen0 (approved):
lgtm

PR #9, opened 2024-07-14
feature: add RPC class

SpectreMercury:
- add RPC Class


Hanssen0:
Methods are added in https://github.com/ckb-ecofund/ccc/commit/7ad54405809269f03bfd613b923f3fe8f897b23e

PR #13, opened 2024-07-22
Add UTXO Global Provider

trongdth:
## What?
- We added code to support the "utxoGlobal" provider, a new Chrome Wallet Extension that runs on the CKB network. More information: [utxoGlobal](https://utxo.global/)

## Why?
- These changes are fully compatible with version 0.0.1 on the Chrome Store.

## Testing?

https://github.com/user-attachments/assets/350bbcd0-28ed-41ba-b7e5-dcd4fffa23fb

## Anything Else?
- The UTXO Global Wallet v0.0.1 does not yet support the "switchNetwork" function. Therefore, you need to switch between Bitcoin or CKB manually.
- The "signTransaction" feature has not been implemented yet. It will be available in the next version and will be updated later on.


trongdth:
[EDITED] 26 Jul 

- The new commit has supported the "switchNetwork" + "signTransaction" functions. This commit is compatible with version 0.0.2 of UTXO Global Wallet Extension (commit https://github.com/UTXO-Global/utxo-wallet-extension/commit/6a6ee495470b3af18ddf878144d5d2143cacf226)
- Version 0.0.2 has not yet been submitted to Chrome Store because version 0.0.1 is still pending.
- The video demo is attached

https://github.com/user-attachments/assets/7de2923a-f02f-4ff1-8a34-461d88a2126f


** Happy coding **

packages/connector/src/connector/index.ts
```diff
@@ -173,7 +173,7 @@ export class WebComponentConnector extends LitElement {
       return;
     }
     this.mainRef.value.style.height = `${
-      this.bodyRef.value?.clientHeight ?? 0
+      this.bodyRef.value?.offsetHeight ?? 0
```
Hanssen0, line 176: Please revert this change

packages/connector/src/signers/index.ts
```diff
@@ -90,6 +91,10 @@ export class SignersController implements ReactiveController {
       this.addSigner("Nostr", "Nostr", NOSTR_SVG, nip07Signer);
     }
 
+    ccc.UtxoGlobal.getUtxoGlobalSigners(client)?.forEach(({ signer, name }) => {
```
Hanssen0, line 94: Since it returns an array, it should always be not null. (But can be empty)

packages/utxo-global/src/btc/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  async getBtcAccount() {
+    const accounts = await this.provider.getAccount();
+    return accounts[0];
+  }
+
+  async getBtcPublicKey(): Promise<ccc.Hex> {
+    const pubKeys = await this.provider.getPublicKey();
+    const account = await this.getBtcAccount();
+    const pubKey = pubKeys.find((_pubKey) => _pubKey.address === account);
+    return ccc.hexFrom(pubKey?.publicKey!);
```
Hanssen0, line 21: It's better to throw an error in unexpected situations when not forced by type explicitly.

packages/utxo-global/src/btc/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  async getBtcAccount() {
+    const accounts = await this.provider.getAccount();
+    return accounts[0];
+  }
+
+  async getBtcPublicKey(): Promise<ccc.Hex> {
+    const pubKeys = await this.provider.getPublicKey();
+    const account = await this.getBtcAccount();
+    const pubKey = pubKeys.find((_pubKey) => _pubKey.address === account);
```
Hanssen0, line 20: _ prefix usually for not used variable but that's not the situation here.

packages/utxo-global/src/ckb/index.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  async getAddressObj(): Promise<ccc.Address | undefined> {
+    const address = await this.getInternalAddress();
+    return await ccc.Address.fromString(address, this.client);
+  }
+
+  async getAddressObjs(): Promise<ccc.Address[]> {
+    const address = await this.getAddressObj();
+
+    if (!!address) {
```
Hanssen0, line 47: if the address is empty, which means the wallet is not correctly connected, we should throw an error instead of acting like normal.

packages/utxo-global/src/ckb/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  ) {
+    super(client);
+  }
+
+  getInternalAddress(): Promise<string> {
+    return this.getAccount();
+  }
+
+  async getIdentity(): Promise<string> {
+    return (await this.getPublicKey()).slice(2);
```
Hanssen0, line 36: Please keep the 0x prefix. Consider extending the SignerCkbPublicKey?

packages/connector/src/signers/index.ts
trongdth: fixed

packages/connector/src/connector/index.ts
trongdth: fixed

packages/utxo-global/src/btc/index.ts
trongdth: fixed

packages/utxo-global/src/btc/index.ts
trongdth: fixed

packages/utxo-global/src/ckb/index.ts
trongdth: fixed

trongdth:
The feedbacks were solved. Could you please re-take a look at it? @Hanssen0

packages/core/package.json
```diff
@@ -60,6 +60,9 @@
     "@joyid/ckb": "^1.0.1",
     "@noble/curves": "^1.4.2",
     "@noble/hashes": "^1.4.0",
+    "@types/secp256k1": "^4.0.6",
```
Hanssen0, line 63: These deps should be removed now.

packages/utxo-global/README.md
```diff
@@ -0,0 +1,45 @@
+<p align="center">
+  <a href="https://ckbccc-demo.vercel.app/">
+    <img alt="Logo" src="https://raw.githubusercontent.com/ckb-ecofund/ccc/master/assets/logo.svg" width="20%" />
+  </a>
+</p>
+
+<h1 align="center" style="font-size: 64px;">
+  CCC's support for Utxo Global
```
Hanssen0, line 8: It's OK to use UTXO in the README if you want.

packages/utxo-global/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@ckb-ccc/utxo-global",
+  "version": "0.0.5-alpha.10",
```
Hanssen0, line 3: Please modify the version to be the same as others.

packages/utxo-global/src/btc/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  async signMessageRaw(message: string | ccc.BytesLike): Promise<string> {
+    const challenge =
+      typeof message === "string" ? message : ccc.hexFrom(message).slice(2);
+    const account = await this.getBtcAccount();
+    return this.provider.signMessage(challenge, account);
+  }
+
+  async signMessage(message: ccc.BytesLike): Promise<ccc.Signature> {
```
Hanssen0, line 44: You can remove this, it has a default implementation.

packages/utxo-global/src/ckb/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  async signMessageRaw(message: string | ccc.BytesLike): Promise<string> {
+    const challenge =
+      typeof message === "string" ? message : ccc.hexFrom(message).slice(2);
+    const account = await this.getAccount();
+    return this.provider.signMessage(challenge, account);
+  }
+
+  async signMessage(message: ccc.BytesLike): Promise<ccc.Signature> {
```
Hanssen0, line 80: You can remove this, it has a default implementation.

And also verifyMessage and verifyMessageRaw below.

packages/utxo-global/src/ckb/index.ts
```diff
@@ -0,0 +1,136 @@
...
+    const pubKey = await this.getPublicKey();
+    return ccc.verifyMessageCkbSecp256k1(message, signature as string, pubKey);
+  }
+
+  async signOnlyTransaction(
+    txLike: ccc.TransactionLike,
+  ): Promise<ccc.Transaction> {
+    const rawTx = cccA.JsonRpcTransformers.transactionFrom(txLike);
+    const txSigned = await this.provider.signTransaction(rawTx);
+    return JSON.parse(txSigned) as ccc.Transaction;
```
Hanssen0, line 121: The ccc.Transaction has extra props like methods, use ccc.Transaction.from instead.

Hanssen0:
Merged https://github.com/ckb-ecofund/ccc/commit/8b7ace26745d8ad21002d97d25bd859888da220c

PR #34, opened 2024-08-29
UTXO Global Wallet / support switchNetwork function

trongdth:


trongdth:
it should be closed coz the same with https://github.com/ckb-ecofund/ccc/pull/25

PR #44, opened 2024-09-11
feat(demo): transfer with timelock

fghdotio:


packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
+"use client";
+
+import React, { useEffect, useState } from "react";
+import { TextInput } from "@/src/components/Input";
+import { Button } from "@/src/components/Button";
+import { Textarea } from "@/src/components/Textarea";
+import { bytesConcat, ccc } from "@ckb-ccc/connector-react";
+import { Since, numFrom, numToBytes } from "@ckb-ccc/core";
```
Hanssen0, line 8: Please import CCC-related methods from the same package and use the `ccc` prefix for all methods to keep consistent.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+    txHash: "0x54090200597d544f0c8e8c02a0dc12715260ccc88dfe097ae76198ba71f71bff",
+    index: 0,
+  },
+  depType: "code",
+}
+const feeRate = 3000;
+
+export default function TimeLockedTransfer() {
+  const { signer, createSender } = useApp();
+  const { log, error } = createSender("Transfer");
```
Hanssen0, line 29: Please update the log title

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+  const handleTimeLockedTransfer = async () => {
+    if (!signer) {
+      return;
+    }
+
+    // Verify destination addresses
+    const toAddress = await ccc.Address.fromString(transferTo, signer.client);
+
+    const currentBlockNumber = await signer.client.getTip();
+    const lockedUntil = new Since("absolute", "blockNumber", currentBlockNumber + numFrom(lockedForBlocks));
```
Hanssen0, line 50: Please use `Since.from` here for more readable code.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+    if (!signer) {
+      return;
+    }
+    // cell ordering is not guaranteed
+    const timeLockedCells: ccc.Cell[] = [];
+    for await (const cell of signer.client.findCells(
+      {
+        script: ccc.Script.from({
+          ...testnetTimeLockScript,
+          args: bytesFromAnyString((await signer.getRecommendedAddressObj()).script.hash()),
```
Hanssen0, line 99: A single signer may have multiple addresses. Use `getAddressObjs` instead.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+        scriptType: "lock",
+        scriptSearchMode: "prefix",
+      },
+      "desc",
+    )) {
+      // Process each cell
+      timeLockedCells.push(cell);
+    }
+    // pick the first one
+    setLiveTimeLockCell(timeLockedCells[0]);
```
Hanssen0, line 110: If you only need the first cell, setting the cell and breaking the loop will be better.
Also, check [Nervos DAO](https://github.com/ckb-ecofund/ccc/blob/master/packages/demo/src/app/connected/(tools)/NervosDao/page.tsx) if you want to list all cells.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  const handleClaim = async () => {
+    if (!signer || !liveTimeLockCell) {
+      return;
+    }
+    const toAddress = await signer.getRecommendedAddressObj();
+    const tx = ccc.Transaction.from({
+      outputs: [{ lock: toAddress.script }],
+      inputs: [ccc.CellInput.from({
```
Hanssen0, line 120: It's not necessary to add a `CellInput.from` here. CCC helps you to do this automatically.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+        previousOutput: liveTimeLockCell.outPoint,
+      })],
+      cellDeps: [ccc.CellDep.from(testnetTimeLockScriptCellDep)],
+      outputsData: [bytesFromAnyString(claimData)],
+    });
+
+    tx.outputs[0].capacity = liveTimeLockCell.cellOutput.capacity;
+
+    await tx.completeInputsByCapacity(signer, 233);
```
Hanssen0, line 129: Please just leave the second args empty.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
fghdotio: Yeah, the first cell is just enough for the current workflow.

packages/demo/src/app/connected/(tools)/TimeLockedTransfer/page.tsx
fghdotio: I believe that explicitly passing the target input cell to be consumed is the safer choice here. For instance, in a claiming scenario, if a user intends to claim an input cell containing 1,000 CKB, they would expect to spend that exact input cell. Allowing CCC to automatically compose inputs could result in selecting two different input cells—for example, one with 500 CKB and another with 600 CKB. Moreover, these two inputs may not be readily available for the user to claim, whereas the original input cell is. The discrepancy in the number of input cells and their lock time could lead to overall confusion.

fghdotio: Sorry for the misunderstanding...

Hanssen0:
Will be merged in #64

PR #52, opened 2024-09-24
add to choose connect type

meitrustadmin:


Hanssen0:
The "redirect" connect type requires a different approach to maintain the application state. Thus, simply changing the connection type won't work.

PR #87, opened 2024-11-19
fix: close&back icon in connector dialog increase 25%

sfsf332:
Trial submission to go through the process. Increase dialog close and back button


Hanssen0 (approved):
LGTM

PR #88, opened 2024-11-26
feat: molecule codec

ashuralyk:
#Description

migrate whole [@ckb-lumos/codec](https://github.com/ckb-js/lumos/tree/develop/packages/codec) package into ccc and replace with this new codec of ccc for native Spore package, which test cases have passed


packages/core/src/barrel.ts
```diff
@@ -9,3 +9,4 @@ export * from "./keystore/index.js";
 export * from "./num/index.js";
 export * from "./signer/index.js";
 export * from "./utils/index.js";
+export * as codec from "./codec/index.js";
```
Hanssen0, line 12: Personally I think `pack` `unpack` `Packed` `Unpacked` `Packable` `Unpackable` are confusing. It looks like `Packed` is always `Bytes` here. Maybe we don't need this much flexibility.

I suggest to use
`encode` `decode` `DataType` instead, and encoder result should always be `Bytes`.
See https://en.wikipedia.org/wiki/Codec#:~:text=A%20codec%20is%20a%20device,a%20portmanteau%20of%20coder%2Fdecoder.

packages/core/src/codec/base.ts
```diff
@@ -0,0 +1,90 @@
...
+export type UnpackParam<T extends AnyCodec> =
+  T extends Codec<any, any, any, infer Unpackable> ? Unpackable : never;
+
+export type Uint8ArrayCodec<Unpacked = any, Packable = Unpacked> = Codec<
+  Uint8Array,
+  Unpacked,
+  Packable
+>;
+
+export type BytesCodec<Unpacked = any, Packable = Unpacked> = Codec<
```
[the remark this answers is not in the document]
ashuralyk: there's a little different, BytesCodec accepts BytesLike as parameter for unpack method, in contrast, Uint8ArrayCodec only accepts Uint8Array

packages/core/src/codec/blockchain.ts
```diff
@@ -0,0 +1,448 @@
...
+  return {
+    ...codec,
+    unpack: (value) => numFrom(codec.unpack(value)).toString(16),
+  };
+}
+
+const HexUint8 = asHexadecimal(Uint8);
+const HexUint32LE = asHexadecimal(Uint32LE);
+const HexUint64LE = asHexadecimal(Uint64LE);
+const HexUint128LE = asHexadecimal(Uint128LE);
```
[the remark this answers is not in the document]
ashuralyk: Hex prefixed types are unpacking to output string, all of these are used in below codes, I'm not sure to delete them.

packages/core/src/codec/base.ts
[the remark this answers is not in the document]
ashuralyk: I can only preserve BytesCodec as the concrete bytes encode/decode, and have every external callers to transfer its BytesLike parameter using ccc.bytesFrom before pack/unpack

packages/core/src/barrel.ts
ashuralyk: confusing indeed there, because of it, I'm concerned to alter it, afraid of causing unexpected problems, however, speaking of which, I'd like to have a try fixing it.

ashuralyk: How about just changing them to:
Packed => Encoded
Packable => Encodable
Unpacked => Decoded
Unpackable => Decodable

packages/core/src/molecule/codec.ts
```diff
@@ -0,0 +1,91 @@
...
+import { bytesFrom, BytesLike } from "../bytes/index.js";
+import {
+  Num,
+  numBeFromBytes,
+  numBeToBytes,
+  numFromBytes,
+  numToBytes,
+} from "../num/index.js";
+
+export interface Codec<Encodable, Decodable = BytesLike> {
```
Hanssen0, line 12:
```typescript
export interface Codec<Encodable, Decoded = Encodable> {
  byteLength?: number;
  encode: (encodable: Encodable) => Bytes;
  decode: (decodable: BytesLike) => Decoded;
}
```

So we can have

```typescript
const TransactionCodec: Codec<ccc.TransactionLike, ccc.Transaction>;
```

How do you think? It seems that the original `Decodable` is never used.

packages/core/src/barrel.ts
```diff
@@ -6,6 +6,7 @@ export * from "./fixedPoint/index.js";
 export * from "./hasher/index.js";
 export * from "./hex/index.js";
 export * from "./keystore/index.js";
+export * as molecule from "./molecule/index.js";
```
Hanssen0, line 9: Maybe `mol` for easier typing?

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+import { bytes, Codec, EncodableType, fixedBytes } from "./codec.js";
+import { Uint32LE } from "./predefined.js";
+
+/**
+ * Vector with fixed size item codec
+ * @param itemCodec fixed-size vector item codec
+ */
+export function fixedItemVec<Encodable>(
+  itemCodec: Codec<Encodable>,
+): Codec<Array<Encodable>> {
```
Hanssen0, line 14: We can use `Encodable[]` here for a more precise type.

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+  if (itemCodec.byteLength === undefined) {
+    throw new Error("fixedItemVec: itemCodec requires a byte length");
+  }
+  return bytes({
+    encode(userDefinedItems) {
+      const encodedArray = userDefinedItems.map((item) =>
+        itemCodec.encode(item),
+      );
+      return bytesConcat(
+        Uint32LE.encode(BigInt(userDefinedItems.length)),
```
Hanssen0, line 24: _See above, this is why Uint32LE should be `Codec<NumLike, Num>`_

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+        encodedArray.reduce(
+          (concated, item) => bytesConcat(concated, item),
+          new ArrayBuffer(0),
```
Hanssen0, line 27: `Bytes` is `Uint8Array`, `new Uint8Array([])` is more precise or `[]` is simpler.

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+export function fixedItemVec<Encodable>(
+  itemCodec: Codec<Encodable>,
+): Codec<Array<Encodable>> {
+  if (itemCodec.byteLength === undefined) {
+    throw new Error("fixedItemVec: itemCodec requires a byte length");
+  }
+  return bytes({
+    encode(userDefinedItems) {
+      const encodedArray = userDefinedItems.map((item) =>
+        itemCodec.encode(item),
```
Hanssen0, line 21: Can we move this into the `reduce` to avoid extra space cost?

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+      const byteLength = 4 + itemCount * itemByteLength;
+      if (value.byteLength !== byteLength) {
+        throw new Error(
+          `fixedItemVec: invalid buffer size, expected ${byteLength}, but got ${value.byteLength}`,
+        );
+      }
+      const decodedArray: Array<Encodable> = [];
+      for (let offset = 0; offset < byteLength; offset += itemByteLength) {
+        decodedArray.push(
+          itemCodec.decode(value.slice(offset, offset + itemByteLength)),
```
Hanssen0, line 49: Can we catch exceptions here and modify the message? e.g. If the itemCodec throws "ItemError: Invalid data", we can wrap it to "FixedItemVec(ItemError: Invalid data)" here for more explicit error message.

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+/**
+ * Option is a dynamic-size type.
+ * Serializing an option depends on whether it is empty or not:
+ * - if it's empty, there is zero bytes (the size is 0).
+ * - if it's not empty, just serialize the inner item (the size is same as the inner item's size).
+ * @param innerCodec
+ */
+export function option<Encodable>(
+  innerCodec: Codec<Encodable>,
+): Codec<Encodable | undefined> {
```
Hanssen0, line 143: `<Encodable | undefined | null, Encodable | undefined>` should be better

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+  return bytes({
+    encode(userDefinedOrNull) {
+      if (userDefinedOrNull === undefined) {
+        return Uint8Array.from([]);
```
Hanssen0, line 147: We can use `bytesFrom` here.

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+ */
+export function table<
+  T extends Record<string, Codec<any>>,
+  R extends object = { [key in keyof T]: EncodableType<T[key]> },
+>(codecLayout: T): Codec<R> {
+  return bytes({
+    encode(object) {
+      const keys = Object.keys(codecLayout);
+      const objectKeys = Object.keys(object);
+      if (JSON.stringify(keys) !== JSON.stringify(objectKeys)) {
```
Hanssen0, line 202: `objectKeys` may have different order with `keys`. Also, `objectKeys` may have more keys than `keys`.

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+      );
+      offsets.push(byteLength);
+      const object = {};
+      for (let i = 0; i < offsets.length - 1; i++) {
+        const start = offsets[i];
+        const end = offsets[i + 1];
+        const field = keys[i];
+        const codec = codecLayout[field];
+        const payload = value.slice(start, end);
+        Object.assign(object, { [field]: codec.decode(payload) });
```
Hanssen0, line 254: Why not just `object[field] = codec.decode(payload);`?

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+ * Struct is a fixed-size type: all fields in struct are fixed-size and it has a fixed quantity of fields.
+ * The size of a struct is the sum of all fields' size.
+ * @param codecLayout a object contains all fields' codec
+ */
+export function struct<
+  T extends Record<string, Codec<any>>,
+  R extends object = { [key in keyof T]: EncodableType<T[key]> },
+>(codecLayout: T): Codec<R> {
+  const codecArray = Object.values(codecLayout);
+  if (codecArray.find((codec) => codec.byteLength === undefined)) {
```
Hanssen0, line 341: `Array.some()`

packages/core/src/molecule/codec.ts
```diff
@@ -0,0 +1,91 @@
...
+export type EncodableType<T extends Codec<any, any>> =
+  T extends Codec<infer Encodable, any> ? Encodable : never;
+export type DecodableType<T extends Codec<any, any>> =
+  T extends Codec<any, infer Decodable> ? Decodable : never;
+
+/**
+ * Create a codec to deal with bytes-like data.
+ * @param codec
+ */
+export function bytes<Encodable>(codec: Codec<Encodable>): Codec<Encodable> {
```
Hanssen0, line 27: It seems that `bytes` and `fixedBytes` are just for type conversion and checking?

packages/core/src/molecule/codec.advanced.ts
```diff
@@ -0,0 +1,373 @@
...
+    throw new Error("struct: all fields must be fixed-size");
+  }
+  return fixedBytes({
+    byteLength: codecArray.reduce((sum, codec) => sum + codec.byteLength!, 0),
+    encode(object) {
+      const keys = Object.keys(codecLayout);
+      const objectKeys = Object.keys(object);
+      if (JSON.stringify(keys) !== JSON.stringify(objectKeys)) {
```
Hanssen0, line 349: Same as `table`

packages/core/src/molecule/codec.advanced.ts
ashuralyk: image

packages/core/src/molecule/codec.advanced.ts
ashuralyk: the packed result won't be undefined, it's just an empty buffer

packages/core/src/molecule/codec.ts
ashuralyk: good pattern, but I'm afraid it's not possible to explicitly declare the Decoded as ccc.Transaction, because the generic Decoded cannot recognize the `from` method in ccc.Transaction, while we cannot directly use `as` to cast ccc.TransactionLike to ccc.Transaction.

so in case of molecule, I think making the Decoded as BytesLike can fit into most of cases.

Hanssen0: > good pattern, but I'm afraid it's not possible to explicitly declare the Decoded as ccc.Transaction, because the generic Decoded cannot recognize the `from` method in ccc.Transaction, while we cannot directly use `as` to cast ccc.TransactionLike to ccc.Transaction.
> 
> so in case of molecule, I think making the Decoded as xxxLike can fit into most of cases, for example, users should manually use ccc.Transaction.from to cast the decoded result of molecule decoders

Maybe we can add a method?
```ts
Codec<Encodable, Decoded>.map<NewDecoded>(transformer: (decoded: Decoded) => NewDecoded): Codec<Encodable, NewDecoded>
```

ashuralyk:
@Hanssen0 already merged "spore searcher" into this PR, if no more problems, this PR could also be merged.

PR #90, opened 2024-12-01
fix: `hexFrom` in `@ckb-ccc/core/barrel` not producing valid hex strings when an empty array is provided

Officeyutong:
```javascript
console.log(hexFrom([])); // outputs "0x"
console.log(hexFrom(new Uint8Array([]))); // outputs "0x"
```        
`hexFrom` will return `0x` which is an invalid hex string when an empty array was provided as argument, since `bytesTo` returns empty string when zero-length bytes was provided.

Illegal hex string like `0x` will be refused by `BigInt` and leads to an exception.

This PR fixes it by providing a default "0" if `bytesTo` returns an empty string.

### DIFF
--- a/packages/core/src/hex/index.ts
+++ b/packages/core/src/hex/index.ts
@@ -26,5 +26,5 @@ export type HexLike = BytesLike;
  * ```
  */
 export function hexFrom(hex: HexLike): Hex {
-  return `0x${bytesTo(bytesFrom(hex), "hex")}`;
+  return `0x${bytesTo(bytesFrom(hex), "hex") || "0"}`;
 }


Hanssen0:
This is the intended behaviour. `hexFrom` is not designed for numeric values but raw bytes data.

For example, if a cell's data is "0x", it stores empty data. But if it's "0x0", it has one bytes of data.

[`ccc.numFrom`](https://docs.ckbccc.com/functions/_ckb_ccc_core.index.ccc.numFrom.html) and [`ccc.numFromBytes`](https://docs.ckbccc.com/functions/_ckb_ccc_core.index.ccc.numFromBytes.html) are designed for transforming bytes/hex number to bigint.

Hanssen0:
Please check if #92 fix `epochFromHex`

Officeyutong:
> Please check if #92 fix `epochFromHex`

It's fixed now, thank you!

PR #105, opened 2024-12-20
feat: @ckb-ccc/ssri and @ckb-ccc/udt release candidate

Alive24:


Hanssen0:
Will be replaced by #130

PR #171, opened 2025-02-28
feat(core): calculate Nervos DAO profit as input capacity

Hanssen0:
This fixes fee calculation for DAO withdrawal txs.
If devs don't change the tx composing logic, sending tx will fail.

### DIFF
--- a/packages/core/src/ckb/transaction.ts
+++ b/packages/core/src/ckb/transaction.ts
@@ -526,6 +526,21 @@ export class CellInput extends mol.Entity.Base<CellInputLike, CellInput>() {
     );
   }
 
+  async getCell(client: Client): Promise<{
+    cellOutput: CellOutput;
+    outputData: Hex;
+  }> {
+    await this.completeExtraInfos(client);
+    if (!this.cellOutput || !this.outputData) {
+      throw new Error("Unable to complete input");
+    }
+
+    return {
+      cellOutput: this.cellOutput,
+      outputData: this.outputData,
+    };
+  }
+
   /**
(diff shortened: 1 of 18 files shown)


packages/core/src/ckb/transaction.ts
```diff
@@ -549,6 +691,15 @@ export class CellInput extends mol.Entity.Base<CellInputLike, CellInput>() {
     this.outputData = cell.outputData;
   }
 
+  /**
+   * The extra capacity created when consume this input.
+   * This is usually NervosDAO interest, see https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0023-dao-deposit-withdraw/0023-dao-deposit-withdraw.md.
+   * And it can also be miners' income. (But this is not implemented yet)
+   */
+  async getExtraCapacity(client: Client): Promise<Num> {
```
ashuralyk, line 699: `extra` here is particularly meant to be the dao profit, which might cause a little confusion, so how about changing its name to `getCapacityProfit` or other more self-explaining alternatives?

ashuralyk (approved):
LGTM, altogether

packages/core/src/ckb/transaction.ts
Hanssen0: As I mentioned, this might also include miners' income in the future, so I think extra capacity might be a better word.

ashuralyk: miner income is also a type of `profit` from my side, anyway, keeping the same is fine

PR #173, opened 2025-03-10
chore: tiny complete for core & spore

ashuralyk:
# Description

for `core`:
while developing `ckb-devrel-indexer`, we encountered an abnormal XUDT format, in which the cell data is empty, according to the RFC of XUDT, we always assume its data is larger than 16 bytes, so that incompatible situation should be considered.

for `spore`:
1. I found there would be a demand that searching spore/cluster under a customized lock script, not only for the signers themself, for example, a customized script locked cluster that can provide on-chain user-defined verification on minting spores.
2. an unusual [version](https://github.com/sporeprotocol/spore-contract/blob/master/docs/VERSIONS.md#tag-021-testnet) of spore in testnet has been applied, so adding this version into predefined config

### DIFF
--- a/.changeset/blue-pans-hope.md
+++ b/.changeset/blue-pans-hope.md
@@ -1,0 +1,5 @@
+---
+"@ckb-ccc/spore": minor
+---
+
+support search spore/cluster under customized lock script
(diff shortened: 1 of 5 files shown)


Hanssen0 (approved):
LGTM

Issue #33, opened 2024-08-28
Add Molecule related features


Hanssen0:
# Background

In the CKB ecosystem, [molecule](https://github.com/nervosnetwork/molecule) is widely used. We initially have [moleculec-es](https://github.com/nervosnetwork/moleculec-es) to generate molecule codecs for JS, but it's written in Golang, so it can not run in a JS runtime.

Later, we have related features in the [Lumos](https://github.com/ckb-js/lumos), see [@ckb-lumos/codec](https://github.com/ckb-js/lumos/tree/develop/packages/codec) and [@ckb-lumos/molecule](https://github.com/ckb-js/lumos/tree/develop/packages/molecule). These two offer compile-time, runtime molecule parsing, and molecule schema defining in JS.

Nowadays, CCC plans to be a successor of Lumos to provide a better developer experience. So, we also want to add the molecule-related feature in CCC.

# Plan

* [ ] Add molecule codecs to `@ckb-ccc/core`. Similar to `@ckb-lumos/codec` but with CCC types.
* [ ] Add a package `@ckb-ccc/molecule` based on codecs. Similar to `@ckb-lumos/molecule`.
* [ ] Decorators to bind codecs to plain TS classes. e.g.
```typescript
@Molecule(codec)
class Script {
...
}
const bytes = Script.from({...}).toBytes();
Script.fromBytes(bytes);
```
This helps developers to use Molecule.

Keith-CY:
@homura will start this issue next week

homura:
I need to postpone addressing this issue because of higher-priority tasks. I'll likely start it next week

homura:
Let me break down the codec feature request. 

In CCC, the code is organized using the OO pattern with many classes such as `Script`, `Transaction`, etc. The request is to enhance any `class` by adding a `static fromBytes` method for the class and a `toBytes` method for its instance.

The enhancer should take any `class` and create a new version of that `class` with the following:

1. A new constructor that can create an enhanced instance

```ts
type Constructor<
  Instance, 
  Arguments extends unknown[] = any[]
> = new (...args: Arguments) => Instance;
```

2. A new prototype method `toBytes`

```ts
type ToBytes<Instance> = Instance & { toBytes(): Bytes }
```

3. A new static method `fromBytes` to create the enhanced instance

```ts
type FromBytes<Class, Instance> = Class & { fromBytes(bytes: Bytes): Instance }
```

Therefore, a `Codec` enhancer can be defined as:

```ts
[...]

PR #174, opened 2025-03-11
feat(mol): add support for fixed-size Union

phroi:
Improve codecs by supporting fixed-size unions

For nervosnetwork/molecule#99, related to #33


packages/core/src/molecule/codec.ts
```diff
@@ -444,8 +445,22 @@ export function union<T extends Record<string, CodecLike<any, any>>>(
   fields?: Record<keyof T, number | undefined | null>,
 ): Codec<UnionEncodable<T>, UnionDecoded<T>> {
   const keys = Object.keys(codecLayout);
+  const values = Object.values(codecLayout);
+  let byteLength = values[0]?.byteLength;
+  for (const { byteLength: l } of values.slice(1)) {
```
Hanssen0, line 450: If the first byteLength is undefined, we can return early to avoid extra checks. Consider to remove the `silce(1)` here.

packages/core/src/molecule/codec.ts
```diff
@@ -444,8 +445,22 @@ export function union<T extends Record<string, CodecLike<any, any>>>(
   fields?: Record<keyof T, number | undefined | null>,
 ): Codec<UnionEncodable<T>, UnionDecoded<T>> {
   const keys = Object.keys(codecLayout);
+  const values = Object.values(codecLayout);
```
Hanssen0, line 448: Consider using `Object.entries()` for performance reasons.

packages/core/src/molecule/signed.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Vector codec for arrays of 8-bit signed integers.
+ */
+export const Int8Vec = vector(Int8);
+
+/**
+ * Creates a codec for encoding and decoding 16-bit signed integers.
+ * @param littleEndian - Indicates whether to use little-endian byte order.
+ * @returns A codec for 16-bit signed integers.
+ */
+function int16Number(littleEndian: boolean): Codec<NumLike, number> {
```
Hanssen0, line 46: I prefer to have sth like `intNumber(byteLength, littleEndian)` to reduce duplicate code, though that means we might not be able to use `DataView`.

`ccc.numToByte` added support for negative numbers. Maybe that helps

https://github.com/ckb-devrel/ccc/pull/177

phroi:
Thank you for your comments, but I'd rather abandon this Pull Request Draft, as its code holds no value:

1. [`ccc.numToByte` support for negative integers](https://github.com/ckb-devrel/ccc/pull/177) is exactly what's needed here, but also keep in mind that it's a foot-gun if used incorrectly: `IntN` will need to carefully check integer bounds to make sure that no two numbers map to the same representation, for example `2^(8*bytes)  - 1` and `-1`. [For example said checks are completely missing from uint](https://github.com/ckb-devrel/ccc/issues/178)

2.  Fixed Union has been proposed, but [its current form has a different implementation than the one implemeted in this PR](https://github.com/nervosnetwork/molecule/issues/99#issuecomment-2723542625) (thanks to implicit padding feature)

Love & Peace, Phroi

phroi:
Reopening this PR: here we'll exclusively just focus on the union codec, adding signed integer codecs is the scope for another PR

phroi:
@Hanssen0 I fixed the small performance issues you noted and added some TypeDocs.

Now comes the most difficult part: [should we allow for implicit padding or not?](https://github.com/nervosnetwork/molecule/issues/99#issuecomment-2724182007) What's your take?

In my opinion I don't really think we should cause it feels kind of a breaking change and makes this a bit more out of reach.

Phroi %59

Hanssen0:
Sorry misclicked!

PR #175, opened 2025-03-12
bug: support customized spore script for transfer&melt method

ashuralyk:
# Description

Reported by UTXOGlobal team that methods `transferSpore` and `meltSpore` are not allowed for user-defined spore scripts. To solve this issue, we should expose `ScriptInfoLike` parameter.

### DIFF
--- a/packages/spore/src/cluster/index.ts
+++ b/packages/spore/src/cluster/index.ts
@@ -145,17 +145,22 @@ export async function transferSporeCluster(params: {
   id: ccc.HexLike;
   to: ccc.ScriptLike;
   tx?: ccc.TransactionLike;
+  scripts?: SporeScriptInfoLike[];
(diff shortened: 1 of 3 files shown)


packages/spore/src/cluster/index.ts
```diff
@@ -145,17 +145,22 @@ export async function transferSporeCluster(params: {
   id: ccc.HexLike;
   to: ccc.ScriptLike;
   tx?: ccc.TransactionLike;
+  scriptInfo?: SporeScriptInfoLike;
```
Hanssen0, line 148: Can we use the same args as `assertCluster` - `scripts` to allow auto choosing scriptInfo?

Hanssen0 (changes requested):

Hanssen0 (approved):
LGTM

PR #202, opened 2025-05-07
feat(core): CKB_TX_MESSAGE_ALL

XuJiandong:
described in: https://github.com/nervosnetwork/rfcs/pull/446

### DIFF
--- a/.changeset/polite-bats-shop.md
+++ b/.changeset/polite-bats-shop.md
@@ -2,4 +2,5 @@
 "@ckb-ccc/core": patch
 ---
 
-Add CKB_TX_MESSAGE_ALL signing encoding
+Add CKB_TX_MESSAGE_ALL: https://github.com/nervosnetwork/rfcs/pull/446
+


XuJiandong:
integration tests: https://github.com/xxuejie/ckb-tx-message-all-test-vector-utils/pull/2

packages/core/src/ckb/transaction.ts
```diff
@@ -1193,6 +1208,83 @@ export class Transaction extends mol.Entity.Base<
...
+
+        if (position === -1) {
+          position = i;
+        }
+      }
+
+      if (position === -1) {
+        return undefined;
+      }
+      if (i == position) {
```
Hanssen0, line 1261: `===` is almost always more recommended than `==` to avoid implicit conversion.

Hanssen0 (commented):
This PR provides a good start for CCC to support `MESSAGE_ALL` signing in the future. However, some API designs need to be discussed more before merging.

packages/core/src/ckb/transaction.ts
```diff
@@ -1137,6 +1137,21 @@ export class Transaction extends mol.Entity.Base<
     hasher.update(raw);
   }
 
+  static hashBytesToHasher(bytes: BytesLike, hasher: Hasher) {
+    const raw = bytesFrom(hexFrom(bytes));
+    hasher.update(numToBytes(raw.length, 4));
+    hasher.update(raw);
+  }
+  static hashBytesOptToHasher(bytes: BytesLike | undefined, hasher: Hasher) {
```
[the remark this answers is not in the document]
XuJiandong: These two functions are helper functions, so it's fine to mark them as private.  
However, there's no need to add them to `Hasher` since they won't be used frequently.

packages/core/src/ckb/transaction.ts
XuJiandong: fixed

XuJiandong:
> I noticed that the logic of hashBytes[Opt]ToHasher can actually be replaced by mol.Bytes[Opt]. I suggest using mol.Bytes[Opt].encode directly to save some code here.

According to the RFC:
> The length of the witness field, packed in little-endian encoded unsigned 32-bit integer.
> The full witness field, or the full content returned by [Load Witness](https://github.com/nervosnetwork/rfcs/blob/bd5d3ff73969bdd2571f804260a538781b45e996/rfcs/0009-vm-syscalls/0009-vm-syscalls.md#load-witness) syscall, given the correct index and source

The hashBytesToHasher function follows the RFC mentioned above and uses the same encoding as Molecule bytes. I believe this version is much cleaner and easier to understand.

The hashBytesOptToHasher function adds an extra header, based on BytesOpt encoding. When BytesOpt is none, it results in zero bytes, so we include an extra header to differentiate it.

Hanssen0 (approved):
LGTM

Issue #204, opened 2025-05-16
molecule some issues


gpBlockchain:
https://github.com/nervosnetwork/ckb-js-vm/issues/71#issuecomment-2885541519

- [x] Uint8Opt.encode(0) = Uint8Array(0) [] ,expected Uint8Array(1) [ 0 ]

```
let ret = mol.Uint8Opt.encode(0)
console.log(ret)

console.log
    Uint8Array(0) []
```

- [x] BoolOpt.encode(false) = Uint8Array(0) [] ,expected Uint8Array(1) [ 0 ]
```
let ret = mol.BoolOpt.encode(false)
        console.log(ret)


console.log
    Uint8Array(0) []
```

- [x] StringOpt.encode("") == Uint8Array(0) [],expected Uint8Array(4) [ 0, 0, 0, 0 ]
```
let ret = mol.StringOpt.encode("")
        console.log(ret)

console.log
    Uint8Array(0) []
```

Hanssen0:
Fix published #206
Will be merged into `master` branch in #207

Hanssen0:
Thanks for your feedback! The issue has been fixed. Also, thanks to @Xujiandong for #208.
Completion of unit tests is planned to reduce the likelihood of this issue.

PR #205, opened 2025-05-16
fix(core): falsy value is not empty in mol.Option

Hanssen0:
#204

### DIFF
--- a/.changeset/cyan-melons-hang.md
+++ b/.changeset/cyan-melons-hang.md
@@ -1,0 +1,5 @@
+---
+"@ckb-ccc/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


XuJiandong:
Is it necessary to check `undefined`?

Hanssen0:
> Is it necessary to check `undefined`?

See [Loose equality using ==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Equality_comparisons_and_sameness#loose_equality_using)

> 2. If one of the operands is null or undefined, the other must also be null or undefined to return true. Otherwise return false.

Nullish check is a special scenario in which `==` is allowed for simplicity.

PR #208, opened 2025-05-16
fix(core): fix molecule option for number zero

XuJiandong:
```js
    let ret = mol.Uint8Opt.encode(0);
    console.log(`ret:${JSON.stringify(ret)}`);
```
The result is `None`(empty) which is unexpected.


Hanssen0:
duplicated #205

Issue #199, opened 2025-05-06
Molecule not support compatible mode


XuJiandong:
In rust, we have `compatible` flag like below:
```rust
    fn from_slice(slice: &'r [u8]) -> VerificationResult<Self> {
        Self::verify(slice, false).map(|_| Self::new_unchecked(slice))
    }
    fn from_compatible_slice(slice: &'r [u8]) -> VerificationResult<Self> {
        Self::verify(slice, true).map(|_| Self::new_unchecked(slice))
    }
```

By default, the ccc molecule sets compatible off.

Hanssen0:
Solved by #209

PR #209, opened 2025-05-17
feat: compatible mode for molecule decode

Alive24:


packages/core/src/molecule/codec.ts
```diff
@@ -18,13 +18,19 @@ import {
 
 export type CodecLike<Encodable, Decoded = Encodable> = {
   readonly encode: (encodable: Encodable) => Bytes;
-  readonly decode: (decodable: BytesLike) => Decoded;
+  readonly decode: (
+    decodable: BytesLike,
+    config?: { compatible?: boolean },
```
Hanssen0, line 23: I thought `isCompatible` should be a good enough name here. But after consideration, I suspect that `compatible` is an accurate and descriptive word here.

If we say `compatible` or `isCompatible`, it sounds more like the `decodable` is `compatible`, while it's not. It's more like we are telling "use the compatible mode", so `isCompatibleMode`  might be more accurate, but still not descriptive enough.

As a result, I suggest describing exactly what we're doing, like `shouldAcceptExtraField`/`isExtraFieldIgnored`. And add a comment here to tell this is the same as `compatible` in Rust.

packages/core/src/molecule/codec.ts
```diff
@@ -371,15 +380,15 @@ export function table<
...
+    decode(buffer, config) {
       const value = bytesFrom(buffer);
       if (value.byteLength < 4) {
         throw new Error(
           `table: too short buffer, expected at least 4 bytes, but got ${value.byteLength}`,
         );
       }
       const byteLength = uint32From(value.slice(0, 4));
-      if (byteLength !== value.byteLength) {
+      if (byteLength !== value.byteLength && !config?.compatible) {
```
Hanssen0, line 391: This doesn't look like the correct implementation. See
https://github.com/nervosnetwork/molecule/blob/e0f6d769ea91fb4a3880019eb5bce7a7975ac650/tools/codegen/src/generator/languages/rust/reader/implementation.rs#L234

"Compatible" means we accept tables with extra fields.

Hanssen0 (changes requested):

packages/core/src/molecule/codec.ts
Alive24: Fixed

packages/core/src/molecule/codec.ts
Alive24: Fixed

packages/core/src/molecule/codec.ts
```diff
@@ -43,7 +49,7 @@ export class Codec<Encodable, Decoded = Encodable> {
         }
         return encoded;
       },
-      (decodable) => {
+      (decodable, config = { isExtraFieldIgnored: false }) => {
```
Hanssen0, line 52: I recommend not setting the default value here. Just leave it undefined. Adding more config fields in the future and forgetting to modify every default value could cause a problem.

packages/core/src/molecule/codec.ts
```diff
@@ -397,9 +429,9 @@ export function table<
         const payload = value.slice(start, end);
         try {
           // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-          Object.assign(object, { [field]: codec.decode(payload) });
-        } catch (e: unknown) {
-          throw new Error(`table.${field}(${e?.toString()})`);
+          Object.assign(object, { [field]: codec.decode(payload, config) });
+        } catch (_e: unknown) {
```
Hanssen0, line 433: I'm curious why we need to change the error name.

Anyway, LGTM

Alive24: It was an accident. Would you please help me revert it if necessary?

Hanssen0 (approved):
LGTM, should solve #199

packages/core/src/molecule/codec.ts
```diff
@@ -18,13 +18,19 @@ import {
 
 export type CodecLike<Encodable, Decoded = Encodable> = {
   readonly encode: (encodable: Encodable) => Bytes;
-  readonly decode: (decodable: BytesLike) => Decoded;
+  readonly decode: (
+    decodable: BytesLike,
+    config?: { isExtraFieldIgnored?: boolean },
```
XuJiandong, line 23: Use term `compatible` which is already used in Rust(https://github.com/search?q=repo%3Anervosnetwork%2Fmolecule%20compatible&type=code) instead of `isExtraFieldIgnored`

XuJiandong (changes requested):

packages/core/src/molecule/codec.ts
Hanssen0: > I thought `isCompatible` should be a good enough name here. But after consideration, I suspect that `compatible` is an accurate and descriptive word here.

> If we say `compatible` or `isCompatible`, it sounds more like the `decodable` is `compatible`, while it's not. It's more like we are telling "use the compatible mode", so `isCompatibleMode`  might be more accurate, but still not descriptive enough.

> As a result, I suggest describing exactly what we're doing, like `shouldAcceptExtraField`/`isExtraFieldIgnored`. And add a comment here to tell this is the same as `compatible` in Rust.

XuJiandong (approved):

PR #210, opened 2025-05-17
feat: @ckb-ccc/molecule

Alive24:
Demo ready and waiting for merge.


packages/molecule/src/utils.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  return sorted;
+}
+
+function id<T>(value: T): T {
+  return value;
+}
+
+type CircularIterator<T> = {
```
Hanssen0, line 201: `CircularIterator` seems to be unused in the code. Consider removing it.

packages/molecule/src/utils.ts
```diff
@@ -0,0 +1,231 @@
...
+interface Node {
+  id: ID;
+  dependencies: ID[];
+}
+
+/**
+ * topological sort with circular check
+ * @param graph
+ */
+export function topologySort<T extends Node>(graph: T[]): T[];
```
Hanssen0, line 147: `topologySort` seems to be unused in the code. Consider removing it.

packages/molecule/src/utils.ts
```diff
@@ -0,0 +1,231 @@
...
+ * mol type `array` and `struct` should have fixed byte length
+ */
+function assertFixedMolType(name: string, map: MolTypeMap): void {
+  const molItem = map[name];
+  nonNull(molItem);
+  const type = molItem.type;
+  switch (type) {
+    case "array": {
+      if (molItem.item !== byte) {
+        assertFixedMolType(molItem.name, map);
```
Hanssen0, line 118: I haven't actually run this code, but it looks like the implementation here will result in an infinite loop. Perhaps we should test for the case where the array item is not `byte`?

Hanssen0 (changes requested):
Thank you for the contribution. However, the style of this code is not good enough to be merged into CCC. Issues include:

1. Inconsistent naming - `getCodecMapFromMol`/`createCodecMap`'s logic are almost always parsing something, but using different names. `nonNull` as a function, it has no verb in its name, and also different will `assert...` functions like `assertFixedMolType`. Moreover, we have `check...` functions that return nothing, basically `assert...` functions.

2. Non-descriptive names - Names like `refs` force devs to look into the code to tell what it is. Others like `toMolTypeMap.results` give us a wrong description - it's an argument, not a result.

3. Performance - This is the least important issue. But still, `MolType[]` is converted to `MolTypeMap` multiple times, and some assertions like `assertFixedMolType` might happen more than once in a single type.

Hanssen0:
I expect the improved version and will review the demo code after that. In the meantime, the first commit modifies `pnpm-lock`, but CI seems to have some problems. Please check.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+      break;
+    }
+    case "union": {
+      // Tuple of [UnionFieldName, UnionFieldId, UnionTypeCodec]
+      const unionCodecs: [string, number, mol.Codec<any, any>][] = [];
+
+      molType.items.forEach((unionTypeItem, index) => {
+        if (unionTypeItem === byte) {
+          unionCodecs.push([unionTypeItem, index, mol.Byte]);
```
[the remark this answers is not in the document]
Alive24: Here when the item doesn't come in (string, number) (which means specifying the field ID), it would use the index as the field ID by default.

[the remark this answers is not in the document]
Hanssen0: I see. The if-else indent is kind of misleading here, would using `else if` be better than `else { if... }`? 

One more thing, if the `unionTypeItem` is `[byte, fieldId]`, what will happen?

It seems that `toCodec` won't handle the case `key === byte`. If completed, I think a lot of the special checks for `byte` can be removed, which will help increase code readability.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,189 @@
...
+    return refs[key];
+  }
+  const molType: MolType = molTypeMap[key];
+  nonNull(molType);
+  let codec: mol.Codec<any, any> | null = null;
+  switch (molType.type) {
+    case "array": {
+      if (molType.name.startsWith("Uint")) {
+        switch (molType.name) {
+          case "Uint8":
```
Hanssen0, line 33: I suggest using codecs record here to keep the code clean. e.g.
```ts
const BUILT_IN_ARRAY_CODECS = {
  Uint8: mol.Uint8,
  Uint16: mol.Uint16,
  ...
  Byte8: mol.Byte8,
  ...
};
...
const builtInCodec = BUILT_IN_ARRAY_CODECS[molType.name];
if (builtInCodec) {
  return builtInCodec;
}
```

Alive24:
> Thank you for the contribution. However, the style of this code is not good enough to be merged into CCC. Issues include:
> 
> 1. Inconsistent naming - `getCodecMapFromMol`/`createCodecMap`'s logic are almost always parsing something, but using different verbs. `nonNull` as a function, it has no verb in its name, and also different from `assert...` functions like `assertFixedMolType`. Moreover, we have `check...` functions that return nothing, basically `assert...` functions.
> 2. Non-descriptive names - Names like `refs` force devs to look into the code to tell what it is. Others like `toMolTypeMap.results` give us a wrong description - it's an argument, not a result.
> 3. Performance - This is the least important issue. But still, `MolType[]` is converted to `MolTypeMap` multiple times, and some assertions like `assertFixedMolType` might happen more than once in a single type.

Names changes are made. Functions are refactored to keep pure and MolTypeDefinition is not converted to MolTypeMap any more and assertions are also improved.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,233 @@
...
+      break;
+    }
+    case "union": {
+      const unionCodecs: [string, number, mol.Codec<any, any>][] = [];
+
+      molTypeDefinition.items.forEach((unionTypeItem, index) => {
+        if (unionTypeItem === BYTE) {
+          unionCodecs.push([unionTypeItem, index, mol.Byte]);
+        } else {
+          if (typeof unionTypeItem === "string") {
```
Hanssen0, line 100: The if-else indent is kind of misleading here, would using else if be better than else { if... }?

One more thing, if the unionTypeItem is [byte, fieldId], what will happen?

It seems that toCodec won't handle the case key === byte. If completed, I think a lot of the special checks for byte can be removed, which will help increase code readability.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,233 @@
...
+ * @throws Error if required dependencies are not found
+ */
+function toCodec(
+  molTypeDefinition: MolTypeDefinition,
+  outputCodecRecord: CodecRecord,
+): mol.Codec<any, any> {
+  if (outputCodecRecord[molTypeDefinition.name]) {
+    return outputCodecRecord[molTypeDefinition.name];
+  }
+  let codec: mol.Codec<any, any> | null = null;
```
Hanssen0, line 37: I suggest `undefined` as the empty value to keep consistency.

packages/molecule/src/type.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export type Table = {
+  type: "table";
+  name: string;
+  fields: Field[];
+};
+
+// mol type definitions
+export type MolTypeDefinition =
```
Hanssen0, line 48: Related to the above comment: Should `BYTE` be a `MolTypeDefinition`? It seems that this should help us reduce the code for special cases.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * @param molTypeDefinitions - The molecule type definitions to convert. Should be obtained by grammar.parse(molSchemaSrc).declares.
+ * @param extraReferences - Optional pre-defined codecs that can be referenced instead of creating new ones
+ * @returns CodecRecord indexed by the type name
+ */
+export function toCodecRecord(
+  molTypeDefinitions: MolTypeDefinition[],
+  extraReferences?: CodecRecord,
+): CodecRecord {
+  const outputCodecRecord: CodecRecord = extraReferences || {};
```
Hanssen0, line 200: `??` is more recommended than `||` to avoid falsy value special cases.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,233 @@
...
+  const processed = new Set<string>();
+
+  function process(definition: MolTypeDefinition) {
+    if (processed.has(definition.name)) return;
```
Hanssen0, line 204: `{ ... }` is always recommended for `if` - too many historical catastrophes have been caused by this.

packages/molecule/src/codec.ts
```diff
@@ -0,0 +1,233 @@
...
+ * @returns CodecRecord indexed by the type name
+ */
+export function toCodecRecord(
+  molTypeDefinitions: MolTypeDefinition[],
+  extraReferences?: CodecRecord,
+): CodecRecord {
+  const outputCodecRecord: CodecRecord = extraReferences || {};
+  const processed = new Set<string>();
+
+  function process(definition: MolTypeDefinition) {
```
Hanssen0, line 203: This function has logic similar to that of the `toCodec` function. Would we let the `toCodec` function do the recursion better?

packages/molecule/src/utils.ts
```diff
@@ -0,0 +1,260 @@
...
+ * - For array and struct types: Ensures they have fixed byte length
+ * - For vector and option types: Validates that their item type exists in either molTypeDefinitions or extraReferences
+ * - For union types: Validates that all union items exist and have valid IDs
+ * - For table types: Validates that all field types exist in either molTypeDefinitions or extraReferences
+ *
+ * @param molTypeDefinitions - Array of mol type definitions to validate
+ * @param extraReferences - Optional record of external codecs that can be used as dependencies
+ * @throws Error if any dependency is not found or if type constraints are violated
+ */
+export function validateMolTypeDefinitions(
```
Hanssen0, line 82: It seems that we always do the validation during the processing step - in the `toCodec`. I'm not sure if it's necessary to pre-validate before that.

packages/molecule/src/utils.ts
```diff
@@ -0,0 +1,260 @@
...
+ * - All fields must be fixed-length types
+ * - If a field is not BYTE, its type must be a fixed-length type
+ *
+ * @param molTypeDefinition - The molecule type definition to validate
+ * @param molTypeDefinitionsAsReferences - Available type definitions that can be referenced
+ * @throws Error if the type is not fixed length or if required dependencies are not found
+ */
+function assertFixedLengthMolType(
+  molTypeDefinition: MolTypeDefinition,
+  molTypeDefinitionsAsReferences: MolTypeDefinition[],
```
Hanssen0, line 193: This name seems way toooooo long for us, making it hard to read. I personally think `AsReferences` is the default behavior and does not need to be mentioned; `AsOutput` seems reasonable (but I hope we don't need to do that).

I suggest just `molDefs` or `molDefsList`, it's clear enough. Readability is always our first consideration.

Hanssen0 (commented):
This version is way better than the last one! Everything is understandable now.

I commented with some questions and suggestions that might help us improve. Please check.

packages/molecule/src/codec.ts
Alive24: 1. Fixed.
2. You're correct. It should handle this situation as you don't have `mol.Byte` in `outputCodecRecord`. Fixed.
3. I'm not sure but it feels like different switch cases here handle byte differently (e.g., as fixed array or dynamic vector of bytes or as one item only in option) while you don't feed `mol.Byte` as a `MolTypeDefinition` to `toCodec`. I'm not sure if you're only referring to Union type as it is the only type that involves a `key` in `codec.ts` while `MolTypeDefinition` is indexed by its `name`.

packages/molecule/src/type.ts
Alive24: It feels a bit weird for me if you insert `BYTE` to `parsedMolTypeDefinitions` during `toCodecRecord` and `parseMoleculeSchema` as you don't put BYTE into the schema. Nonetheless, we can do so if you prefer less code; it's ok to me if we don't since its definition comes fundamentally from molecule itself instead of user schema.

packages/molecule/src/codec.ts
Alive24: Here I'm trying to keep it pure; I tried to do so but I think that using `toCodec` for recursion would need to pass outputCodecRecord around and look similar to this as well unless you want `toCodec` also returns the dependended together with the original one. You might want to propose a specific change?

packages/molecule/src/utils.ts
Alive24: It checks the length for fixed length mol type too, and I don't think `toCodec` would do this.

packages/molecule/src/utils.ts
Alive24: I would prefer molTypeDefinitions as I don't like short form personally. Proposed new names.

packages/molecule/src/utils.ts
Alive24: And this one checks dependencies across MolTypeDefinitions before actually building the codec record while `toCodec` checks dependencies across entries in the CodecRecord.

packages/molecule/src/utils.ts
Hanssen0: `molTypeDedinitions` doesn't feel right to me. It is too similar to the previous name (only a "s"), hiding possible mistakes.

Hanssen0: Additional note:
`molTypeDefinitions(List)` doesn't actually provide any information to me: it's basically the same as its type.

It's acceptable, but if we can name it under the context of this function, it might be better.

packages/molecule/src/codec.ts
Hanssen0: 1. I don't think this way works - We're making sth pure by making another thing impure - and even worse, the `process` is a local function/closure holding reference to a local variable.
2. We still need to pass `outputCodecRecord` around to read.

Now I have why `toCodec` can't be a pure function - we ignored the topological sorting step, so it's possible that `toCodec` parse a def before its sub defs.

I don't think introducing topological sort is necessary since it brings extra complexity. Then, I guess it's impossible/hard to keep this step pure, so just make it impure.

Alive24:
Commit history seems imreparable. Started a new one at #212

Issue #227, opened 2025-06-14
[FR] Merge iCKB Utils UDTHandler features into CCC's Udt


phroi:
Hello CKB DevRel folks!!
I was wondering, have you considered merging iCKB Utils' UDTHandler Features into the CCC library's Udt?

## Summary

We currently maintain three parallel UDT implementations:

1. CCC's `Udt` class (metadata, `transfer`, `mint`, basic `completeBy`)  
2. iCKB Utils' `UdtHandler`/`UdtManager` (SmartTransaction-aware balance tracking, input aggregation, change-cell via `completeUdt`)  
3. iCKB Core's `IckbUdtManager` (overrides input balance for iCKB UDT cells, receipt and deposit transformation logic)

This duplication leads to inconsistent behaviors, extra maintenance, and diverging sUDT data-length checks.

## CCC Feature Request

In the CCC library's `Udt` class, add methods/properties from iCKB Utils' UdtManager:
- `getInputsUdtBalance(client, tx)`  (open to change the name)
- `getOutputsUdtBalance(tx)` (we can find a better name) 
- `completeBy(tx, signer, options?)`, support `shouldAddInputs?: boolean` and `compressState?: boolean` as options
- cellDeps: ccc.CellDep[];

**Note**: I'm open to change any of the previous names if necessary.

## ICKB side of things

Deprecate iCKB Utils' standalone `UdtHandler`/`UdtManager`. Instead, let `IckbUdtManager` extend the enhanced CCC `Udt` class, overriding only:

- The xUDT data-length check (≥16 bytes on-chain) 
- Custom input-balance logic for:
  - iCKB receipt cells
  - DAO deposit/withdrawal cells
- Accept both `ccc.Transaction` and `SmartTransaction` as `completeBy` inputs, detect type, use `smartTransaction.copy(tx)` when needed in SmartTransaction `completeFee`, and return the same transaction type.  

Keep up the Great Work, Phroi %28

Hanssen0:
Thank you for your suggestions! It's a reasonable feature to implement in CCC for better code reuse.

Before starting, I wonder what the `compressState` option is?

For the `cellDeps` property, if we add an `onTransactionCreated` hook for manipulating `Transaction`, how do you think?

phroi:
> I wonder what the compressState option is?

If `shouldAddInputs` is evaluated to `true` and at least one additional UDT cell is required, **add all** UDT cells to compress state rent. Feel free to change the name!

[...]

PR #228, opened 2025-06-14
feat(udt): balance related util methods

Hanssen0:
For #227

### DIFF
--- a/packages/udt/src/udt/index.ts
+++ b/packages/udt/src/udt/index.ts
@@ -2,6 +2,194 @@ import { ccc } from "@ckb-ccc/core";
 import { ssri } from "@ckb-ccc/ssri";
 
 /**
+ * Error thrown when there are insufficient UDT coins to complete a transaction.
+ * This error provides detailed information about the shortfall, including the
+ * exact amount needed, the UDT type script, and an optional custom reason.
+ *
+ * @public
+ * @category Error
+ * @category UDT
+ *
+ * @example
+ * ```typescript
+ * // This error is typically thrown automatically by UDT methods
+ * try {
+ *   await udt.completeInputsByBalance(tx, signer);
+ * } catch (error) {
+ *   if (error instanceof ErrorUdtInsufficientCoin) {
+ *     console.log(`Error: ${error.message}`);
+ *     console.log(`Shortfall: ${error.amount} UDT tokens`);
+ *     console.log(`UDT type script: ${error.type.toHex()}`);
+ *   }
+ * }
+ * ```
+ */
+export class ErrorUdtInsufficientCoin extends Error {
+  /**
+   * The amount of UDT coins that are insufficient (shortfall amount).
+   * This represents how many more UDT tokens are needed to complete the operation.
+   */
+  public readonly amount: ccc.Num;
+
+  /**
+   * The type script of the UDT that has insufficient balance.
+   * This identifies which specific UDT token is lacking sufficient funds.
+   */
+  public readonly type: ccc.Script;
+
+  /**
+   * Creates a new ErrorUdtInsufficientCoin instance.
+   *
+   * @param info - Configuration object for the error
+   * @param info.amount - The amount of UDT coins that are insufficient (shortfall amount)
+   * @param info.type - The type script of the UDT that has insufficient balance
+   * @param info.reason - Optional custom reason message. If not provided, a default message will be generated
+   *
+   * @example
+   * ```typescript
+   * // Manual creation (typically not needed as the error is thrown automatically)
+   * const error = new ErrorUdtInsufficientCoin({
+   *   amount: ccc.numFrom(1000),
+   *   type: udtScript,
+   *   reason: "Custom insufficient balance message"
+   * });
+   *
+   * // More commonly, catch the error when it's thrown by UDT methods
+   * try {
+   *   const result = await udt.completeInputsByBalance(tx, signer);
+   * } catch (error) {
+   *   if (error instanceof ErrorUdtInsufficientCoin) {
+   *     // Handle the insufficient balance error
+   *     console.error(`Insufficient UDT: need ${error.amount} more tokens`);
+   *   }
+   * }
+   * ```
+   *
+   * @remarks
+   * The error message format depends on whether a custom reason is provided:
+   * - With custom reason: "Insufficient coin, {custom reason}"
+   * - Without custom reason: "Insufficient coin, need {amount} extra coin"
+   */
+  constructor(info: {
+    amount: ccc.NumLike;
+    type: ccc.ScriptLike;
+    reason?: string;
+  }) {
+    const amount = ccc.numFrom(info.amount);
+    const type = ccc.Script.from(info.type);
+    super(`Insufficient coin, ${info.reason ?? `need ${amount} extra coin`}`);
+    this.amount = amount;
+    this.type = type;
+  }
+}
+
+/**
+ * Configuration object type for UDT instances.
+ * This type defines the optional configuration parameters that can be passed
+ * when creating a UDT instance to customize its behavior.
+ *
+ * @public
+ * @category Configuration
+ * @category UDT
+ */
+export type UdtConfigLike = {
+  /**
+   * Optional SSRI executor instance for advanced UDT operations.
+   * When provided, enables SSRI-compliant features like metadata queries
+   * and advanced transfer operations.
+   */
+  executor?: ssri.Executor | null;
+
+  /**
+   * Optional custom search filter for finding UDT cells.
+   * If not provided, a default filter will be created that matches
+   * cells with the UDT's type script and valid output data length.
+   */
+  filter?: ccc.ClientIndexerSearchKeyFilterLike | null;
+};
+
+/**
+ * Configuration class for UDT instances.
+ * This class provides a structured way to handle UDT configuration parameters
+ * and includes factory methods for creating instances from configuration-like objects.
+ *
+ * @public
+ * @category Configuration
+ * @category UDT
+ *
+ * @example
+ * ```typescript
+ * // Create configuration with executor
+ * const config = new UdtConfig(ssriExecutor);
+ *
+ * // Create configuration with both executor and filter
+ * const config = new UdtConfig(
+ *   ssriExecutor,
+ *   ccc.ClientIndexerSearchKeyFilter.from({
+ *     script: udtScript,
+ *     outputDataLenRange: [16, 32]
+ *   })
+ * );
+ *
+ * // Create from configuration-like object
+ * const config = UdtConfig.from({
+ *   executor: ssriExecutor,
+ *   filter: { script: udtScript, outputDataLenRange: [16, "0xffffffff"] }
+ * });
+ * ```
+ */
+export class UdtConfig {
+  /**
+   * Creates a new UdtConfig instance.
+   *
+   * @param executor - Optional SSRI executor for advanced UDT operations
+   * @param filter - Optional search filter for finding UDT cells
+   */
+  constructor(
+    public readonly executor?: ssri.Executor,
+    public readonly filter?: ccc.ClientIndexerSearchKeyFilter,
+  ) {}
+
+  /**
+   * Creates a UdtConfig instance from a configuration-like object.
+   * This factory method provides a convenient way to create UdtConfig instances
+   * from plain objects, automatically converting filter-like objects to proper
+   * ClientIndexerSearchKeyFilter instances.
+   *
+   * @param configLike - Configuration-like object containing executor and/or filter
+   * @returns A new UdtConfig instance with the specified configuration
+   *
+   * @example
+   * ```typescript
+   * // Create from object with executor only
+   * const config = UdtConfig.from({ executor: ssriExecutor });
+   *
+   * // Create from object with filter only
+   * const config = UdtConfig.from({
+   *   filter: {
+   *     script: udtScript,
+   *     outputDataLenRange: [16, "0xffffffff"]
+   *   }
+   * });
+   *
+   * // Create from object with both
+   * const config = UdtConfig.from({
+   *   executor: ssriExecutor,
+   *   filter: { script: udtScript, outputDataLenRange: [16, 32] }
+   * });
+   * ```
+   */
+  static from(configLike: UdtConfigLike) {
+    return new UdtConfig(
+      configLike.executor ?? undefined,
+      configLike.filter
+        ? ccc.ClientIndexerSearchKeyFilter.from(configLike.filter)
+        : undefined,
+    );
+  }
+}
+
+/**
  * Represents a User Defined Token (UDT) script compliant with the SSRI protocol.
  *
  * This class provides a comprehensive implementation for interacting with User Defined Tokens,
@@ -13,35 +201,144 @@ import { ssri } from "@ckb-ccc/ssri";
  * @category Token
  */
 export class Udt extends ssri.Trait {
+  /**
+   * The type script that uniquely identifies this UDT token.
+   * This script is used to distinguish UDT cells from other cell types and
+   * to identify which cells belong to this specific UDT token.
+   *
+   * @remarks
+   * The script contains:
+   * - `codeHash`: Hash of the UDT script code
+   * - `hashType`: How the code hash should be interpreted ("type" or "data")
+   * - `args`: Arguments that make this UDT unique (often contains token-specific data)
+   *
+   * @example
+   * ```typescript
+   * const udt = new Udt(codeOutPoint, scriptConfig);
+   * console.log(`UDT script hash: ${udt.script.hash()}`);
+   * console.log(`UDT args: ${udt.script.args}`);
+   *
+   * // Check if a cell belongs to this UDT
+   * const isUdt = udt.isUdt(cell);
+   * ```
+   */
   public readonly script: ccc.Script;
 
   /**
-   * Constructs a new UDT (User Defined Token) script instance.
-   * By default it is a SSRI-compliant UDT. By providing `xudtType`, it is compatible with the legacy xUDT.
+   * The search filter used to find UDT cells controlled by signers.
+   * This filter is automatically configured to match cells with this UDT's type script
+   * and appropriate output data length (minimum 16 bytes for UDT balance storage).
    *
-   * @param executor - The SSRI executor instance.
-   * @param code - The script code cell of the UDT.
-   * @param script - The type script of the UDT.
+   * @remarks
+   * The filter includes:
+   * - `script`: Set to this UDT's type script
+   * - `outputDataLenRange`: [16, "0xffffffff"] to ensure valid UDT cells
+   *
+   * This filter is used internally by methods like:
+   * - `calculateInfo()` and `calculateBalance()` for scanning all UDT cells
+   * - `completeInputs()` and related methods for finding suitable input cells
+   *
    * @example
    * ```typescript
-   * const udt = new Udt(executor, code, script);
+   * const udt = new Udt(codeOutPoint, scriptConfig);
+   *
+   * // The filter is used internally, but you can access it if needed
+   * console.log(`Filter script: ${udt.filter.script?.hash()}`);
+   * console.log(`Output data range: ${udt.filter.outputDataLenRange}`);
+   *
+   * // Manually find cells using the same filter
+   * for await (const cell of signer.findCells(udt.filter)) {
+   *   console.log(`Found UDT cell with balance: ${ccc.udtBalanceFrom(cell.outputData)}`);
+   * }
    * ```
+   */
+  public readonly filter: ccc.ClientIndexerSearchKeyFilter;
+
+  /**
+   * Constructs a new UDT (User Defined Token) script instance.
+   * By default it is a SSRI-compliant UDT. This class supports both SSRI-compliant UDTs and legacy sUDT/xUDT standard tokens.
+   *
+   * @param code - The script code cell outpoint of the UDT. This points to the cell containing the UDT script code
+   * @param script - The type script of the UDT that uniquely identifies this token
+   * @param config - Optional configuration object for advanced settings
+   * @param config.executor - The SSRI executor instance for advanced UDT operations. If provided, enables SSRI-compliant features
+   * @param config.filter - Custom search filter for finding UDT cells. If not provided, a default filter will be created
+   *
+   * @example
+   * ```typescript
+   * // Basic UDT instance
+   * const udt = new Udt(
+   *   { txHash: "0x...", index: 0 }, // code outpoint
+   *   { codeHash: "0x...", hashType: "type", args: "0x..." } // type script
+   * );
+   *
+   * // UDT with SSRI executor for advanced features
+   * const ssriUdt = new Udt(
+   *   codeOutPoint,
+   *   typeScript,
+   *   { executor: ssriExecutor }
+   * );
+   *
+   * // UDT with custom filter (advanced usage)
+   * const customUdt = new Udt(
+   *   codeOutPoint,
+   *   typeScript,
+   *   {
+   *     filter: {
+   *       script: typeScript,
+   *       outputDataLenRange: [16, 32], // Only cells with 16-32 bytes output data
+   *     }
+   *   }
+   * );
+   * ```
+   *
+   * @remarks
+   * **Default Filter Behavior:**
+   * If no custom filter is provided, a default filter is created with:
+   * - `script`: Set to the provided UDT type script
+   * - `outputDataLenRange`: [16, "0xffffffff"] to match valid UDT cells
+   *
+   * **SSRI Compliance:**
+   * When an executor is provided, the UDT instance can use SSRI-compliant features like:
+   * - Advanced transfer operations
+   * - Metadata queries (name, symbol, decimals, icon)
+   * - Custom UDT logic execution
+   *
+   * **Legacy Support:**
+   * Even without an executor, the UDT class supports basic operations for legacy sUDT/xUDT tokens.
    */
   constructor(
     code: ccc.OutPointLike,
     script: ccc.ScriptLike,
-    config?: {
-      executor?: ssri.Executor | null;
-    } | null,
+    config?: UdtConfigLike | null,
   ) {
     super(code, config?.executor);
     this.script = ccc.Script.from(script);
+    this.filter = ccc.ClientIndexerSearchKeyFilter.from(
+      config?.filter ?? {
+        script: this.script,
+        outputDataLenRange: [16, "0xffffffff"],
+      },
+    );
   }
 
   /**
    * Retrieves the human-readable name of the User Defined Token.
+   * This method queries the UDT script to get the token's display name,
+   * which is typically used in user interfaces and wallets.
    *
-   * @returns A promise resolving to the token's name.
+   * @param context - Optional script execution context for additional parameters
+   * @returns A promise resolving to an ExecutorResponse containing the token's name,
+   *          or undefined if the name is not available or the script doesn't support this method
+   *
+   * @example
+   * ```typescript
+   * const udt = new Udt(codeOutPoint, scriptConfig);
+   * const nameResponse = await udt.name();
+   * if (nameResponse.res) {
+   *   console.log(`Token name: ${nameResponse.res}`);
+   * }
+   * ```
    */
   async name(
     context?: ssri.ContextScript,
@@ -60,8 +357,22 @@ export class Udt extends ssri.Trait {
   }
 
   /**
-   * Retrieves the symbol of the UDT.
-   * @returns The symbol of the UDT.
+   * Retrieves the symbol (ticker) of the User Defined Token.
+   * The symbol is typically a short abbreviation used to identify the token,
+   * similar to stock ticker symbols (e.g., "BTC", "ETH", "USDT").
+   *
+   * @param context - Optional script execution context for additional parameters
+   * @returns A promise resolving to an ExecutorResponse containing the token's symbol,
+   *          or undefined if the symbol is not available or the script doesn't support this method
+   *
+   * @example
+   * ```typescript
+   * const udt = new Udt(codeOutPoint, scriptConfig);
+   * const symbolResponse = await udt.symbol();
+   * if (symbolResponse.res) {
+   *   console.log(`Token symbol: ${symbolResponse.res}`);
+   * }
+   * ```
    */
   async symbol(
     context?: ssri.ContextScript,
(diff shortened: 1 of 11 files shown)


packages/udt/src/udt/index.ts
```diff
@@ -274,33 +367,514 @@ export class Udt extends ssri.Trait {
...
+   * const isValid = udt.isUdtCell(cellOutput, outputData);
+   * console.log(`Is valid UDT cell: ${isValid}`); // true
+   * ```
+   *
+   * @remarks
+   * The method checks two conditions:
+   * 1. The cell's type script matches this UDT's script
+   * 2. The output data is at least 16 bytes long (required for UDT balance storage)
+   */
+  isUdtCell(cellOutputLike: ccc.CellOutputLike, outputData: ccc.HexLike) {
```
phroi, line 394: Why not using [CellLike](https://api.ckbccc.com/types/_ckb_ccc_core.index.ccc.CellLike.html)?

packages/udt/src/udt/index.ts
```diff
@@ -274,33 +367,514 @@ export class Udt extends ssri.Trait {
...
+   *
+   * const inputBalance = await udt.getInputsBalance(tx, client);
+   * console.log(`Total UDT input balance: ${inputBalance}`);
+   * ```
+   *
+   * @remarks
+   * This method only counts inputs that have the same type script as this UDT instance.
+   * Inputs without a type script or with different type scripts are ignored.
+   */
+  async getInputsBalance(
```
phroi, line 423: Why not also returning the CKB occupied by said UDT cells?

packages/udt/src/udt/index.ts
```diff
@@ -274,33 +367,514 @@ export class Udt extends ssri.Trait {
...
+   * const outputBalance = udt.getOutputsBalance(tx);
+   * console.log(`Total UDT output balance: ${outputBalance}`); // 1500
+   * ```
+   *
+   * @remarks
+   * This method only counts outputs that have the same type script as this UDT instance.
+   * Outputs without a type script or with different type scripts are ignored.
+   * This is a synchronous method since output data is already available in the transaction.
+   */
+  getOutputsBalance(txLike: ccc.TransactionLike): ccc.Num {
```
phroi, line 473: For future SSRI compatibility we should make it async

packages/udt/src/udt/index.ts
```diff
@@ -274,33 +367,514 @@ export class Udt extends ssri.Trait {
...
+   * ```
+   *
+   * @remarks
+   * This method is useful for:
+   * - Validating transaction balance conservation
+   * - Calculating how much UDT is being destroyed in burn operations
+   * - Detecting minting operations (negative burned balance)
+   * - Ensuring sufficient UDT inputs are provided for transfers
+   */
+  async getBalanceBurned(
```
phroi, line 516: I would call it getBalanceDelta, cause this delta may indeed be positive or negative

phroi: Also if we add the CKB to the input and output methods, also this one should return its delta

packages/udt/src/udt/index.ts
```diff
@@ -274,33 +367,514 @@ export class Udt extends ssri.Trait {
...
+      (await this.getBalanceBurned(tx, from.client)) -
+      ccc.numFrom(balanceTweak ?? 0);
+    const capacityBurned =
+      (await tx.getFee(from.client)) - ccc.numFrom(capacityTweak ?? 0);
+
+    if (balanceBurned >= ccc.Zero && capacityBurned >= ccc.Zero) {
+      return 0;
+    }
+
+    const { addedCount, accumulated } = await tx.completeInputs(
```
phroi, line 582: This is difficult to override, it could make sense to partition this in multiple overridable methods

phroi: For example it could make sense to have a generator methods that returns said cells wrapped, possibly wrapped into an object where each cell is annotated with its ckb and udt value

packages/udt/src/udt/index.ts
Hanssen0: Good idea!

phroi:
> If we want to collect all UDT cells, I personally think [Transaction.completeInputsAll](https://api.ckbccc.com/classes/_ckb_ccc_core.index.ccc.Transaction.html#completeInputsAll) will be better (of course, we can create a wrapper for it in the Udt class).

I'm not sure. While it is in line with the current Transaction methods, I don't particularly like it.  To me seems orthogonal to what the other methods offer, so adding as an options seems natural.

What's your reason for making it into a method?

Phroi %51

packages/udt/src/udt/index.ts
Hanssen0: Once in a while, I name it `balanceDiff`. I think delta has the same problem as diff, in that we can't intuitively understand whether it represents input minus output or vice versa.

Hanssen0:
> What's your reason for making it into a method?

Good question. In short, the single responsibility principle and fewer options will make the method more stable.

From the implementation perspective, unlike `shouldAddInputs`, which is for disabling existing logic, adding `compressState` requires extra logic in the `completeFee` method.

I also agree that it's orthogonal to methods like `completeFee` or `Udt.complete`, but I think that's the reason why we should **not** couple it with these methods. This adds complexity to the overall system for limited benefit (saving a line of code, or maybe even less, since the extra options also require code).

Hanssen0:
> the single responsibility principle and fewer options will make the method more stable.

Further read: https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/
In-depth discussion of the problems that such parameters may cause.

packages/udt/src/udt/index.ts
phroi: You make a fair point, it's not a given this entities have a Outpoint. In my implementation it did, cause it was only used on on-chain cells

packages/udt/src/udt/index.ts
phroi: Delta usually is defined Output minus Input, differently from diff, which has no such connotation. That said, Burned is surely more colorful and probably intuitive, maybe it'll cause also more raised eyebrows, but nothing a good TypeDoc can't fix

phroi:
> Added getInputsInfo and getOutputsInfo and wrapped them to get...Balance for convenience.

Good idea, on one side we have `get...Info` (used mainly internally and overridable) and on the other `get...Balance` (which filters out unneeded information for normal use cases), great change!! :+1::+1:

Hanssen0:
> Let's make an example, I don't like `findCells` vs `findCellsOnChain`. I would add `onChain : bool` as I did in my implementations. Your current stance explain that choice.

Yes. One exception is [`Client.findTransactions`](https://github.com/ckb-devrel/ccc/blob/813319a4201528647b4d8b61e3d57997582edd06/packages/core/src/client/client.ts#L429) because we want to keep the method design mapped to the RPC calls. I really regret this choice, I mean, look at the method signature!!!

We've been adding more unit tests recently, and splitting methods into smaller units helps us write shorter tests (and makes them easier for others to override).

packages/udt/src/udt/index.ts
phroi: Yeah, you're right, my underlying issue is with `tx.completeInputs`: it may be a fine piece of engineering, I don't like it, my bad!! :rofl:

phroi:
> One exception is [`Client.findTransactions`](https://github.com/ckb-devrel/ccc/blob/813319a4201528647b4d8b61e3d57997582edd06/packages/core/src/client/client.ts#L429) because we want to keep the method design mapped to the RPC calls. I really regret this choice, I mean, look at the method signature!!!

My eyes are burning!!! :rofl:

packages/udt/src/udt/index.ts
```diff
@@ -274,33 +372,631 @@ export class Udt extends ssri.Trait {
...
+    const tx = ccc.Transaction.from(txLike);
+    const balanceBurned =
+      (await this.getBalanceBurned(tx, from.client)) -
+      ccc.numFrom(balanceTweak ?? 0);
+    const capacityBurned =
+      (await tx.getFee(from.client)) - ccc.numFrom(capacityTweak ?? 0);
+
+    if (balanceBurned >= ccc.Zero && capacityBurned >= ccc.Zero) {
+      return 0;
+    }
```
phroi, line 702: This little piece of code is efficiently creating more UDT cells out of thin air: while we could easily compress the state by checking the UDT cells capacity flow (similar to the original @rink1969 idea), we instead are checking the global transaction capacity flow.

I'd propose to evaluate:

```
  const [inUdt, inCkb] = await this.getInputInfo();
  const [outUdt, outCkb] = await this.getOutputInfo();

  const udtBurned = inUdt - outUdt -  ccc.numFrom(balanceTweak ?? 0);// Why not a Num in the first place?
  const ckbBurned =  inUdt - outUdt - ccc.numFrom(capacityTweak ?? 0);// Why not a Num in the first place?

    if (udtBurned >= ccc.Zero && ckbBurned >= ccc.Zero) {
      return 0;
    }
```

phroi:
> will it be better to add the extra logic to Udt.complete by inheriting?

For example, if we split `completeInputs` and we split `tx.completeInputs` into its own internal method and we move the cellDeps hook to its own method, then completeInputs is a general method, that does the following: get initial balance, call internal completeInputs .... and call add cell deps hook.

What's your take on this?

phroi:
Also, another details, how many usecases outside of UDT we have for `ccc.ErrorTransactionInsufficientCoin`?

If all the use-cases are exclusively in UDT, why not bring that Error class definition over here and add a proper message formatting so that it can actually be displayed into an user interface?

So using ccc.FixedPontToString(amount, decimals) and mentioning the symbol of the UDT causing the error. Here we are in UDT so it seems appropriate, differently from Core Transaction.

Whats your take on this?

packages/udt/src/udt/index.ts
Hanssen0: I can live with both, there is no difference in complexity or efficiency.

Hanssen0:
> If all the use-cases are exclusively in UDT, why not bring that Error class definition over here and add a proper message formatting so that it can actually be displayed into an user interface?

It's reasonable to move the Error to `Udt`, I'll handle this.

> So using ccc.FixedPontToString(amount, decimals) and mentioning the symbol of the UDT causing the error.

Considering the extra cost (SSRI call or Indexer query), this might not. I think we should let the application do this.

phroi:
> Considering the extra cost (SSRI call or Indexer query), this might not.

Reasonable

> It's reasonable to move the Error to Udt

Not sure TBH, if you are not willing to add more metadata, it may may make more sense where it is currently.

I guess that I'll consider sub-classing that error to add more metadata in my iCKB subclass :thinking:

Hanssen0:
> Not sure TBH, if you are not willing to add more metadata, it may may make more sense where it is currently.

After we deprecated the `Transaction.completeInputsByUdt`, nothing in the `Transaction` uses `ErrorTransactionInsufficientCoin`. So it do should be moved.

> I wonder, what if instead we make the error a bit more general and optionally let devs specify those details in sub-classes of UDT?

What do you think if we add a `reason?: string` property to allow richer information? If devs want to add more metadata/properties, they can inherit the Error class.

packages/udt/src/udt/index.ts
Hanssen0: e.g. If we have ten 100 UDT cells with 140 CKB for each

For a meta transaction A
```
Inputs: -
Outputs:
  150 UDT in 140 CKB
```
 
After `Udt.completeBy`, we have
Transaction A1
```
Inputs:
  2 * 100 UDT in 140 CKB
Outputs:
  150 UDT in 140 CKB
  50 UDT in 140 CKB
```
 
This is the same for both implementations.
 
But for Transaction B
```
Inputs: -
Outputs:
  150 UDT in 140 CKB
  280 CKB
```
 
After the current `Udt.completeBy`, we have
 Transaction B1
 ```
 Inputs:
   4 * 100 UDT in 140 CKB
 Outputs:
   150 UDT in 140 CKB
   280 CKB
   250 UDT in 140 CKB
 ```
 
If we only consider the UDT capacity flow, it will be
Transaction B2
```
Inputs:
  2 * 100 UDT in 140 CKB
Outputs:
  150 UDT in 140 CKB
  280 CKB
  50 UDT in 140 CKB
```

Obviousily B1 compressed more UDT than B2.

phroi:
Reviewing the newest version now

phroi:
I was wondering about non-SSRI UDTs, I don't see any handling in the main UDT class, so I was wondering:

- How does it work for  for methods like `name`?
- How does the fallback kicks in?
- Is there a database somewhere?

Hanssen0:
> How are devs supposed to get the UDT user balance?

Do you think getting balance from the explorer API is a good idea?

packages/udt/src/udt/index.ts
phroi: I'm not really sure that it makes much of a difference, cause your current algo already works pretty well

That said, if there is a chance to add more UDT cells, with this change they get added without extra RPC calls (except for the possible extra `get_cells` call)

phroi:
> Do you think getting balance from the explorer API is a good idea?

I don't like it, so I wouldn't push it on the devs. Others may like it and so for them is ok

packages/udt/src/udt/index.ts
Hanssen0: It's kinda edge case if devs added some plain CKB cells along with UDT cells. Though I don't know when it will happen, it's a scenario.

Hanssen0:
> > Do you think getting balance from the explorer API is a good idea?
> 
> I don't like it, so I wouldn't push it on the devs. Others may like it and so for them is ok

I see. I will add an method for accumulating UDT balance.

packages/udt/src/udt/index.ts
phroi: For example, I was thinking about NervosDAO withdrawals, like iCKB, and in the same tx the user fan out his UDT holdings

Hanssen0: Anyway, it won't increase the complexity, so it is worth.

phroi:
Going to sleep, thank you for the your time and availability! Wish you a nice evening :hugs:

Phroi %67

Hanssen0:
Methods added:
`calculateInfo` `calculateBalance` `completeInputsAll`
Edge case fixed for `completeInputsByBalance`

phroi:
Starting the next round of review

phroi:
I was wondering, UDT constructor takes specifically `Code`, excluding `DepGroup`, what are the reasons behind this restriction?

While I understand that this is for compatibility with SSRI, it creates unnecessary friction for non-SSRI UDTs.

phroi:
Glad to see the implementation of `calculateInfo`/`calculateBalance`! A couple of observations:

- You previously mentioned an explorer API: if it exists, please include it as one of the first TypeDoc comments, so devs can make an informed choice.
- This is a naive implementation, whatever happens a new exhaustive `get_cells` call will be made (which, given our stance on compressing state, should be fine as likely only one cell will be fetched)
- If by any chance client local cache is out of sync, the result will be incorrect.

My usual way to locally estimate UDT balance would be:
- Find the cell set all cells while extending them with their UDT value, letting in the way the user choose if he want to use under the hood `findCells` or `findCellsOnChain`
- Balance is easy to calculate using the cell set
- User can choose to add all cells or let `completeX` do its job with the warmed cache

What are your considerations on this?

phroi:
I mean, generally I like that the final dev is not directly touching the UDT cells, so in this way we can abstract away from him the exact shape of these cells. Just I wonder if there is a less naive way of validly approaching the problem

phroi:
Since you are the implementer, I was wondering, in which way do you forecast devs will sub-class UDT to support additional cells (that while not being themselves UDTs) have an UDT value?

Which methods would you suggest to said devs to override?

Phroi %51

Hanssen0:
> I was wondering, UDT constructor takes specifically `Code`, excluding `DepGroup`, what are the reasons behind this restriction?
> 
> While I understand that this is for compatibility with SSRI, it creates unnecessary friction for non-SSRI UDTs.

I think `code` here is different from `cellDep`. Yes, the main reason here is for SSRI, or any other method that reads information from the script code. It's more like the preimage of `code_hash`.

Indeed, if a script is not SSRI-compatible, this introduces extra complexity. If we change the `code` to `OutPointLike | CellDepLike | CellDepLike[]`, how do you think?

P.S.: I personally don't like `depGroup`, as it introduces too much complexity and implicitness without bringing a corresponding benefit. It is very disconnected from the design of the entire transaction structure.

Hanssen0:
> You previously mentioned an explorer API: if it exists, please include it as one of the first TypeDoc comments, so devs can make an informed choice.

I'm not sure if it's a good idea to include it here. Other platforms, such as Nervos Docs or CCC Docs, may be more suitable.

> * If by any chance client local cache is out of sync, the result will be incorrect.
> My usual way to locally estimate UDT balance would be:
>
> * Find the cell set all cells while extending them with their UDT value, letting in the way the user choose if he want to use under the hood findCells or findCellsOnChain
> * Balance is easy to calculate using the cell set

Indeed. I'll add an option for choosing onchain or not, and onchain should be the default.
For the later implementation, though the time cost is unavoidable, I prefer to keep the space cost constant.

> I mean, generally I like that the final dev is not directly touching the UDT cells, so in this way we can abstract away from him the exact shape of these cells.

Yes and no. I think [mercury](https://github.com/nervosnetwork/mercury) is an attempt on this road, but soon we realize the cost is much more than the gain.

> Since you are the implementer, I was wondering, in which way do you forecast devs will sub-class UDT to support additional cells (that while not being themselves UDTs) have an UDT value?

I have no idea. The original design here is for sUDT-like scripts, so it's "one script - one token". We want to make its code reusable as much as possible, but if this leads to over-engineering, we will not consider it for the time being.

> Back to CellDeps, I noticed that they are added them only in mint and transfer. Is there a good reason for not adding them in completeX methods?

It makes sense, I think we should add it.

phroi:
If you don't mind, one side question: why did you define `ccc.Zero`?

At the beginning I adopted it broadly in my code, then I realize it may introduce more trouble than its worth, so I removed all references :thinking:

phroi:
By the way, nice to see that you made `UDT.filter` configurable by the dev, I just wanted to suggest that!

phroi:
Good call on `addCellDeps(txLike: ccc.TransactionLike): ccc.Transaction`: perfect balance between achieving your goal in the simplest way, while letting external devs achieve theirs by sub-classing `UDT.addCellDeps` :+1::+1:

Hanssen0:
> If you don't mind, one side question: why did you define `ccc.Zero`?
> 
> At the beginning I adopted it broadly in my code, then I realize it may introduce more trouble than its worth, so I removed all references 🤔

For the balance between readability and compatibility. Other options we have
1. `ccc.numFrom(0)`: It's just too long. Although I might still use it sometimes.
2. `BigInt(0)`: It is not semantically correct. We are using `Num` type instead of `BigInt` type.
3. `0n`: It's great if it doesn't involve compatibility issues. Sometimes we can't use it.

phroi:
> `0n`: It's great if it doesn't involve compatibility issues. Sometimes we can't use it.

GPT agrees with your stance:

> The `0n` literal (the BigInt zero) only works in runtimes that support ES2020+ BigInts. If your TS is compiled down to ES5 or you need to run in older browsers or Node versions without BigInt, 0n will error or be stripped out. In those cases you might fallback to a global Zero (e.g. BigInt(0)) that you control and can polyfill.
>
> No TS-level “compatibility flag” lets you emit 0n for older targets, you either target ES2020+ or supply your own shim.

On my side, I long moved on to ES2020 and all of its niceties. While I understand CCC stance to maximize compatibility, I wonder,  how safe are gonna be those **non**-ES2020+ platforms for running any kind of crypto stuff? 🙀

phroi:
To be honest (and we already discussed about this a few days ago), I don't like `isUdtCell` signature, cause it will cause trouble if in the future we (or a subclasser) want to expand the scope of UDT class to include more protocols, so using UDT in the broader sense of User Defined Token.

I would change the code from:

```typescript
  isUdtCell(cellOutputLike: ccc.CellOutputLike, outputData: ccc.HexLike) {
    return (
      (ccc.CellOutput.from(cellOutputLike).type?.eq(this.script) ?? false) &&
      ccc.bytesFrom(outputData).length >= 16
    );
  }
```

to:

```typescript
  isUdtCell(cellLike : {cellOutput: ccc.CellOutputLike, outputData: ccc.HexLike}) {
    return (
      (ccc.CellOutput.from(cellLike.cellOutput).type?.eq(this.script) ?? false) &&
      ccc.bytesFrom(cellLike.outputData).length >= 16
    );
  }
```

Hanssen0:
> > `0n`: It's great if it doesn't involve compatibility issues. Sometimes we can't use it.
> 
> GPT agrees with your stance:
> 
> > The `0n` literal (the BigInt zero) only works in runtimes that support ES2020+ BigInts. If your TS is compiled down to ES5 or you need to run in older browsers or Node versions without BigInt, 0n will error or be stripped out. In those cases you might fallback to a global Zero (e.g. BigInt(0)) that you control and can polyfill.
> > No TS-level “compatibility flag” lets you emit 0n for older targets, you either target ES2020+ or supply your own shim.
> 
> On my side, I long moved on to ES2020 and all of its niceties. While I understand CCC stance to maximize compatibility, I wonder, how safe are gonna be those **non**-ES2020+ platforms for running any kind of crypto stuff? 🙀

No idea. I don't suggest using non-ES2020 anymore, but I can't decide for devs.

> ``` ts
>  isUdt(cellLike : {cellOutput: ccc.CellOutputLike, outputData: ccc.HexLike}) {
> ```

It looks better!

Hanssen0:
> On the same line of toughs as `isUDT`, I was wondering if it would make more sense to transform calls to `ccc.udtBalanceFrom` into an ovveridable method and same with a second method that creates the UDT change cell.
> 
> What's your take on this?

While this makes sense, I don't think it will be very helpful in the foreseeable future. A script that is inconsistent with the current data layout will cause a lot of problems, both in the on-chain scripts and in the application. I think we should not expect the `Udt` class to handle all of these things, which is a bit over-designed.

phroi:
> > Since you are the implementer, I was wondering, in which way do you forecast devs will sub-class UDT to support additional cells (that while not being themselves UDTs) have an UDT value?
> 
> I have no idea. The original design here is for sUDT-like scripts, so it's "one script - one token". We want to make its code reusable as much as possible, but if this leads to over-engineering, we will not consider it for the time being.

I gave some thoughts about this and I was thinking about keeping the iCKB UDT subclass simple and overriding just `addCellDeps` and `getInputsInfo`. (I'll know for sure only when coding tho)

It will still create some trouble as the `calculateBalance` result will always be lower or equal than the user real UDT worth (which includes iCKB receipts), but should be ok.

What's your take on this?

phroi:
> Maybe overriding the `calculateBalance` method?

If I override `calculateBalance` then I need to override also `completeInputs` method to automatically fetch Receipts cells. At that point I need to override that hypothetical `ccc.udtBalanceFrom` method we just discussed before, cause iCKB Receipts are not UDT cells.

Would you reconsider your [previous stance](https://github.com/ckb-devrel/ccc/pull/228#issuecomment-2995299225)?

phroi:
> I don’t quite know the structure of the iCKB recipient cell. If you can provide some relevant documentation, I might be able to understand it more easily

[Here you go](https://github.com/ickb/whitepaper?tab=readme-ov-file#deposit-phase-1), feel free to ask away any doubt! :hugs: 

> So what you mean is that the recipe cell contains a portion of the amount that can be converted into iCKB, but this portion of the amount is not stored in the first 16 bytes?

That's correct, to be precise, there is no way in which it could be calculated due to the transaction header not being accessible during transaction validation in deposit phase 1. If said header was accessible, then I would directly mint UDTs :wink: 

> how to generate the change cell

[That would be a normal UDT change cell](https://github.com/ickb/whitepaper?tab=readme-ov-file#deposit-phase-2), inclusive the iCKB Value of Receipt

> what to do if we want to apply the change to an existing cell

Two meanings, so I'll analyze both:
- **Input Receipt**: Receipts can only be converted, never updated
-  **Output Receipt**: Receipts are added in output only when making iCKB deposits. If you need to account for an additional iCKB deposit I would suggest just adding an additional Receipt. And that's how I do it.

Does this help clarifying why cells that are not UDTs can still hold UDT value? Any doubts?

phroi:
@Hanssen0 what kind of changes would you like to make to this PR?

Hanssen0:
> @Hanssen0 what kind of changes would you like to make to this PR?

I want to merge this PR as soon as we can. It has been staying here for too long and contains modifications/updates that affect other things. I understand your concerns about SSRI types, and they are reasonable. However, given that they have already been included in previous releases & predictably will take some time to implement, I suggest that we should discuss this further after completing other tasks & merging them.

The remaining issues are what you mentioned before: that some UDT-like cells might not store their data in the first 16 bytes. It would be nice if we could make the `udtBalanceFrom` method easier to override. I'm still unsure how to proceed due to my limited knowledge about what we aim to achieve with these cells. It would be great if you could propose some possible solutions.

Hanssen0:
/canary
For testing

phroi:
**Note for future readers**: in my opinion this PR should not be merged as it stands.

This PR has the following issues:
1. There is no method for accessing the UDT value of a single input/output cell.
2. There is no way to cleanly account for cells that have UDT value, but are not UDT cells.
3. UDT is strongly tied to SSRI, but it shouldn't, as UDT is independent from SSRI.
4. UDT doesn't allow DepGroups as its CellDeps.
5. UDT Metadata behind async methods is source of unnecessary troubles.
6. `completeInputs` only compatible cached cells, not directly fetched from `on chain`.
7. Closures make the understanding of some methods more difficult for devs.

That's why [I proposed another WIP PR](https://github.com/ckb-devrel/ccc/pull/250), which addresses the aforementioned issues.

> It would be great if you could propose some possible solutions.

I already proposed, I will review here 🤗

>  It [...] contains modifications/updates that affect other things

Yep, wouldn't it make more sense to merge those separately?

>  I understand your concerns about SSRI types, and they are reasonable.

The more work we put towards them, the more they are difficult to remove later on. Right?

For example, what if a non compatible SSRI-alike comes later on? And then another and another

>  some UDT-like cells might not store their data in the first 16 bytes. It would be nice if we could make the udtBalanceFrom method easier to override.

Yep, you may want to take a look at [how `infoFrom` fills exactly that role in my PR](https://github.com/ckb-devrel/ccc/pull/250#discussion_r2277577306)

About the integration of this class in iCKB, as [I was mentioning some time ago](https://github.com/ckb-devrel/ccc/pull/228#issuecomment-2995308235):

> I gave some thoughts about this and I was thinking about keeping the iCKB UDT subclass simple and overriding just addCellDeps and getInputsInfo. (I'll know for sure only when coding tho)
> 
> It will still create some trouble as the calculateBalance result will always be lower or equal than the user real UDT worth (which includes iCKB receipts), but should be ok.

That said, currently I'm puzzled by the lack of a method to get the UDT value of a single  input/output Cell.

How do you propose to address this issue @Hanssen0?

Love & Peace, Phroi %85

Hanssen0:
> Yep, wouldn't it make more sense to merge those separately?

Agree. I should divide this PR into smaller pieces so that we can merge the less controversial content.

> The more work we put towards them, the more they are difficult to remove later on. Right?

Yes. But I would say this PR is not making UDT more SSRI - it's already coupled with SSRI. We can still do some optimization on top of this, as long as it doesn't make things worse.

> Yep, you may want to take a look at [how `infoFrom` fills exactly that role in my PR](https://github.com/ckb-devrel/ccc/pull/250#discussion_r2277577306)

Now I see more clearly about why `Info` existed there, not only for performance, but also for abstracting UDT infos. However, the additional thing that requires devs to maintain is still a problem. Let's see if there's a way to avoid it.

Hanssen0:
> This PR has the following issues:
> 1. There is no method for accessing the UDT value of a single input/output cell.
> 2. There is no way to cleanly account for cells that have UDT value, but are not UDT cells.
> 3. UDT is strongly tied to SSRI, but it shouldn't, as UDT is independent from SSRI.
> 4. UDT doesn't allow DepGroups as its CellDeps.
> 5. UDT Metadata behind async methods is source of unnecessary troubles.
> 6. `completeInputs` only compatible cached cells, not directly fetched from `on chain`.
> 7. Closures make the understanding of some methods more difficult for devs.

Most of them are not *issues of this PR*, but rather issues that existed in the current codebase, and should not be expected to be all solved once this PR is merged.

Hanssen0:
Please review these three PRs first. Merging them should make this PR less messy:

https://github.com/ckb-devrel/ccc/pull/258
https://github.com/ckb-devrel/ccc/pull/259
https://github.com/ckb-devrel/ccc/pull/260

Hanssen0:
> Good work there, I review them and they are perfect!!
> 
> Phroi %63

Awesome! I think now we can begin to break this PR into smaller pieces. Here's my plan:

1. Let's focus on the `complete*` methods first. On this stage, only ordinary UDT cells will be considered. We should be able to move every UDT-related method out of `Transaction` after this.
2. Try to add more querying methods, like calculating info of a `Signer`. Meanwhile, we should have methods for parsing a single UDT cell. On this stage, we should consider if these new methods can be overridden to accommodate some unconventional UDT cell requirements.
3. Allow more than one cell deps, and also dep groups.
4. Decouple SSRI with UDT.

This PR will focus on the first stage. After we finish that, we can have new PRs for the following.

How do you think?

Hanssen0:
> I reviewed the current state of UDT and it is reasonable, just surprised to see `calculateInfo` there!

You suggested that before, and I'm happy to add that. If devs don't need to maintain them manually, it does help.

I've trimmed down a little bit of the code and removed the `calculateInfo` temporary (will add them back in the next PR). How do you think?

For the `complete*` methods, I recommend devs to use
* `completeInputsByBalance`
* `completeInputsAll`
* `completeChangeToOutput`
* `completeChangeToLock`
* `completeBy` (This should be most used)
for normal cases. Only use
* `completeInputs`
* `complete`
if the above methods can't satisfy what they need.

phroi:
> How do you think?

GPT-4o says:
> The phrase "How do you think?" is not commonly used in English. A more natural way to ask for someone's opinion or perspective would be "What do you think?" or "How do you feel about this?"

I'm really sorry, no disrespect intended, my English has room for improvement too :hugs:

> I've trimmed down a little bit of the code and removed the calculateInfo temporary (will add them back in the next PR). How do you think?

Everything seems reasonable, I meant exactly that about `calculateInfo`

Just one thing that makes me wonder is `filter`: correct if we handle only UDT, but we'll be able to extend it later?

I know that it's tightly coupled with tx.completeInputs

Maybe it's not necessary for it to be extendable, cause maybe we don't want to automatically add non-UDT cells

Maybe a second filter or something for other cells types will be needed by subclasses and that's about it

Phroi %17

phroi:
[I replied in the phase 2 of UDT PR](https://github.com/ckb-devrel/ccc/pull/261#issuecomment-3193903729) :hugs:

PR #235, opened 2025-07-03
feat: add SignerCkbMultisig module and dependent features

ashuralyk:
# Description

Consider we have multiple versions of multiple signature contract deployed on CKB network, and moreover, construction of multiple signature transaction is relatively complicated, so it's indeed necessary to implement a new signer, which is `SignerCkbMultisig`, to provide an easy-to-use approach for multisig.

### DIFF
--- a/packages/core/src/signer/ckb/signerCkbMultisigReadonly.ts
+++ b/packages/core/src/signer/ckb/signerCkbMultisigReadonly.ts
@@ -198,40 +198,41 @@ export class SignerCkbMultisigReadonly extends Signer {
     txLike: TransactionLike,
   ): Promise<{ script: Script; cellDeps: CellDepInfo[] }[]> {
     const tx = Transaction.from(txLike);
-    const availableMultisigs = await this.getAddressObjs();
+    const multisigs = await this.getAddressObjs();
+    if (multisigs.length === 0) {
+      return [];
+    }
 
     const scripts: { script: Script; cellDeps: CellDepInfo[] }[] = [];
-    for (const multisig of availableMultisigs) {
-      for (const input of tx.inputs) {
-        const {
-          cellOutput: { lock },
-        } = await input.getCell(this.client);
+    for (const input of tx.inputs) {
+      const {
+        cellOutput: { lock },
+      } = await input.getCell(this.client);
 
-        if (scripts.some(({ script }) => script.eq(lock))) {
-          continue;
-        }
+      if (scripts.some(({ script }) => script.eq(lock))) {
+        continue;
+      }
 
-        if (lock.eq(multisig.script)) {
-          const scriptInfo = await this.client.findKnownScript(lock);
-          if (scriptInfo) {
-            scripts.push({
-              script: lock,
-              cellDeps: scriptInfo.cellDeps,
-            });
-          } else {
-            if (typeof this.multisigInfo.knownMultisigScript === "string") {
-              // Generally, this branch could not be reached
-              throw new Error(
-                `Unsupported multisig script: ${this.multisigInfo.knownMultisigScript}`,
-              );
-            }
-            scripts.push({
-              script: lock,
-              cellDeps: this.multisigInfo.knownMultisigScript.cellDeps.map(
-                (cellDep) => CellDepInfo.from(cellDep),
-              ),
-            });
+      if (multisigs.some(({ script }) => script.eq(lock))) {
+        const scriptInfo = await this.client.findKnownScript(lock);
+        if (scriptInfo) {
+          scripts.push({
+            script: lock,
+            cellDeps: scriptInfo.cellDeps,
+          });
+        } else {
+          if (typeof this.multisigInfo.knownMultisigScript === "string") {
+            // Generally, this branch could not be reached
+            throw new Error(
+              `Unsupported multisig script: ${this.multisigInfo.knownMultisigScript}`,
+            );
           }
+          scripts.push({
+            script: lock,
+            cellDeps: this.multisigInfo.knownMultisigScript.cellDeps.map(
+              (cellDep) => CellDepInfo.from(cellDep),
+            ),
+          });
         }
       }
     }


packages/examples/src/transferFromMultisig.ts
```diff
@@ -0,0 +1,85 @@
...
+
+// Charles checks if the transaction is fulfilled
+if (await multisigSigners[2].signaturesFulfilled(tx)) {
+  console.log("Charles: signature is fulfilled");
+} else {
+  console.log("Charles: signature is not fulfilled");
+}
+
+// Charles sends the transaction
+const txHash = await multisigSigners[2].sendTransaction(tx);
```
Hanssen0, line 84: If the signatures were not fulfilled, can we add a readable error message to `sendTransaction` to help locate the problem?

packages/examples/src/transferToMultisig.ts
```diff
@@ -0,0 +1,51 @@
...
+const threshold = 2;
+const firstNthMustMatch = 0;
+
+const publicKeys = signers.map((signer) => signer.publicKey);
+const metadata = ccc.multisigMetadataFromPubkeys(
+  publicKeys,
+  threshold,
+  firstNthMustMatch,
+);
+const multisigLockScript = await ccc.Script.fromKnownMultisigScript(
```
Hanssen0, line 27: Is it possible to use `SignerCkbMultisig` to help construct such an address/script? Maybe we can have a `SignerCkbMultisigReadonly` and let `SignerCkbMultisig` extend it.

packages/core/src/ckb/transaction.ts
```diff
@@ -1595,6 +1595,40 @@ export class Transaction extends mol.Entity.Base<
...
+   * @param threshold - The threshold of the multisig script.
+   * @param client - The client for complete extra infos in the transaction.
+   * @returns A promise that resolves to the prepared transaction
+   *
+   * @example
+   * ```typescript
+   * await tx.prepareMultisigWitness(script, multisigMetadata, threshold, client);
+   * ```
+   */
+  async prepareMultisigWitness(
```
Hanssen0, line 1612: It appears that this logic is closely coupled to the Multisig Signer. Maybe moving it to the Signer implementation is better than making it in the `Transaction`

packages/core/src/ckb/script.ts
```diff
@@ -97,6 +98,57 @@ export function hashTypeFromBytes(bytes: BytesLike): HashType {
...
+ * ```typescript
+ * const { metadata, scriptArgs } = multisigMetadataFromPubkeys({
+ *   pubkeys: ["0x1234...", "0x5678...", "0x90ab...", "0xabcd..."],
+ *   threshold: 2,
+ *   mustMatch: 1,
+ * });
+ * ```
+ */
+
+export function multisigMetadataFromPubkeys(
```
Hanssen0, line 120: I suggest moving to the `signerCkbMultisig.ts` for better cohesion.

packages/core/src/ckb/script.ts
```diff
@@ -205,10 +257,57 @@ export class Script extends mol.Entity.Base<ScriptLike, Script>() {
...
+   *   {
+   *     relative: "absolute",
+   *     metric: "blockNumber",
+   *     value: 1000,
+   *   }
+   * );
+   * ```
    */
+
+  static async fromKnownMultisigScript(
```
Hanssen0, line 287: I suggest moving to the signerCkbMultisig.ts for better cohesion.

packages/core/src/signer/ckb/signerCkbMultisig.ts
```diff
@@ -0,0 +1,245 @@
...
+import { CellDepInfo, Client, KnownScript } from "../../client/index.js";
+import { Hex, hexConcat, hexFrom, HexLike } from "../../hex/index.js";
+import { SignerCkbPrivateKey } from "./signerCkbPrivateKey.js";
+
+export interface MultisigInfo {
+  pubkeys: HexLike[];
+  threshold: number;
+  mustMatch: number;
+  since?: SinceLike;
+  multisigScript?:
```
Hanssen0, line 18: Is it possible to let the `Signer` support both script versions at the same time? e.g., we can let the `Signer` return two addresses for `getAddresses`. Similar to https://github.com/ckb-devrel/ccc/blob/master/packages/core/src/signer/evm/signerEvm.ts

packages/core/src/hex/index.ts
```diff
@@ -28,3 +28,19 @@ export type HexLike = BytesLike;
...
+ *
+ * @param hexLikes - The HexLike values to concatenate.
+ * @returns A Hex string representing the concatenated values.
+ *
+ * @example
+ * ```typescript
+ * const hexString = hexConcat("0x68656c6c6f", "0x776f726c64"); // Outputs "0x68656c6c6f776f726c64"
+ * ```
+ */
+export function hexConcat(...hexLikes: HexLike[]): Hex {
```
Hanssen0, line 44: Great! This method simplifies hex manipulations.

packages/core/src/signer/ckb/signerCkbMultisig.ts
```diff
@@ -0,0 +1,245 @@
...
+  SinceLike,
+  Transaction,
+  TransactionLike,
+} from "../../ckb/index.js";
+import { CellDepInfo, Client, KnownScript } from "../../client/index.js";
+import { Hex, hexConcat, hexFrom, HexLike } from "../../hex/index.js";
+import { SignerCkbPrivateKey } from "./signerCkbPrivateKey.js";
+
+export interface MultisigInfo {
+  pubkeys: HexLike[];
```
Hanssen0, line 14: `MultisigInfo` was used as `Signer.multisig` member's type, so I don't think we should use `*Like` here. Considering adding a `MulsigInfoLike` type for arg types.

Hanssen0 (commented):
Thanks for your contribution! This PR implemented the multisig feature correctly and also added some useful utilities. There are still some improvements that can be made to the interface. Please check.

packages/core/src/ckb/script.ts
```diff
@@ -97,6 +98,57 @@ export function hashTypeFromBytes(bytes: BytesLike): HashType {
...
+  const pubkeyBlake160Hashes = pubkeys.map((pubkey) =>
+    hashCkb(hexFrom(pubkey)).slice(0, 42),
+  );
+  return hexConcat(
+    "0x00",
+    hexFrom(mustMatch.toString(16)),
+    hexFrom(threshold.toString(16)),
+    hexFrom(pubkeyBlake160Hashes.length.toString(16)),
+    ...pubkeyBlake160Hashes,
+  );
```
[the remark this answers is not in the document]
Hanssen0: I think using `numToBytes` will be better here, to improve readability and avoid confusion.

packages/core/src/signer/ckb/signerCkbMultisig.ts
```diff
@@ -0,0 +1,245 @@
...
+      // Signatures array is placed after the multisig metadata, in 65 bytes per signature
+      const signatures =
+        witness.lock
+          .slice(metadata.length)
+          .match(/.{1,130}/g)
+          ?.map(hexFrom) || [];
+      const insertIndex = signatures.findIndex((sig) => sig === emptySignature);
+      if (signatures.length < this.multisig.threshold) {
+        throw new Error("Not enough signature slots to threshold");
+      }
```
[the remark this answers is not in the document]
Hanssen0: Not sure if it's necessary. Does the script fail if the signatures exceed the threshold?

packages/examples/src/transferFromMultisig.ts
ashuralyk: `sendTransaction` through `signerCkbMultisig` will automatically check the fulfillment state and throw error if not fit all of requirement threshold.

packages/core/src/signer/ckb/signerCkbMultisigReadonly.ts
```diff
@@ -0,0 +1,320 @@
...
+   *
+   * @param txLike - The transaction to prepare.
+   * @param scriptLike - The script to prepare.
+   * @returns A promise that resolves to the prepared transaction
+   */
+  async prepareTxMultisigWitness(
+    txLike: TransactionLike,
+    scriptLike: ScriptLike,
+  ) {
+    const tx = Transaction.from(txLike);
```
Hanssen0, line 221: If `txLike` is not an instance of `Transaction`, this will cause a new instance to be created, and all modifications here won't be able to affect the original argument.

Hanssen0 (changes requested):
This PR provides a functional multisig signer, which is great. However, considering its lack of support for possible devnet/unknown scripts, it's more like a proof of concept than a production-ready tool.

Further design is needed.

Hanssen0:
Refactored in #349

PR #241, opened 2025-07-30
feat: add margin feature to enable zero-transfer fee

ashuralyk:
# Description
Current spore sdk implementation didn't achieve a zero-transfer-fee feature when transfering spore, this feature should come up with a little change on low-level transaction fee completion logic


Hanssen0 (changes requested):
I don't think injecting this logic into `Transaction.complete` is a good idea. This makes things over-complicated. Do we have other ways to implement the pay fee by margin feature?

PR #244, opened 2025-08-07
Simplify MapLru, while improving Complexity

### LINKED ISSUE #243: `MapLru` has `O(n)` Complexity for `get`, `set`, and `delete`
Hey folks at CKB DevRel!!

The [current implementation of `MapLru` uses an internal array (`lru`)](https://github.com/ckb-devrel/ccc/blob/7ff50feb133e86fa4512cb5c8fa784d08d04a610/packages/core/src/client/cache/memory.advanced.ts#L145-L190) to track the least recently used keys. This approach leads to performance issues, particularly with larger caches due to  **`O(n)` Complexity for `get`, `set`, and `delete`**.

The `get`, `set`, and `delete` operations require traversing the `lru` array to find and potentially remove/re-insert the key. The `indexOf` and `splice` operations on arrays have a time complexity of O(n), where n is the number of elements in the array. For large caches, this linear search and modification become a bottleneck.

## Solutions

Drop array. Just use the native map order and O(1) operations complexity, so just delete and re-add an entry to signify its recency.

Alternatively, consider using a Doubly Linked List instead of the array to keep track of insertion order. These structures can provide O(1) complexity for LRU operations.

## Impact

Addressing this inefficiency will improve the overall performance of the cache and ensure more accurate LRU behavior. This is especially important for applications where the cache is heavily used.

Love & Peace, Phroi %89

phroi:
Closes #243

### DIFF
--- a/packages/core/src/client/cache/memory.advanced.ts
+++ b/packages/core/src/client/cache/memory.advanced.ts
@@ -142,49 +142,73 @@ export function filterCell(
   return true;
 }
 
-export class MapLru<K, V> extends Map<K, V> {
-  private readonly lru: K[] = [];
-
+/**
+ * A Least Recently Used (LRU) cache implemented using a Map.
+ *
+ * This class extends the built-in Map to provide an LRU cache with a fixed capacity.
+ * When the cache is full, the least recently used entry is automatically evicted.
+ *
+ * @template K The type of the keys in the cache.
+ * @template V The type of the values in the cache. Values are guaranteed to be non-nullable.
+ */
+export class MapLru<K, V> extends Map<K, NonNullable<V>> {
+  /**
+   * Constructs a new MapLru instance.
+   *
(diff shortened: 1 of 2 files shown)


packages/core/src/client/cache/memory.advanced.ts
```diff
@@ -142,49 +142,74 @@ export function filterCell(
...
-      this.delete(this.lru[0]);
-      this.lru.shift();
-    }
-
-    return val;
-  }
 
-  set(key: K, value: V) {
-    this.get(key);
+    const value = super.get(key)!;
```
Hanssen0, line 182:
```ts
const map = new Map<number, number | undefined | null>();

console.log(map.has(1), map.has(2));
map.set(1, undefined);
map.set(2, null);
console.log(map.has(1), map.has(2), Array.from(map.keys()));
```

We got
```
false false
true true [1, 2]
```

So I don't think we can safely assert the value is non-null

PR #248, opened 2025-08-08
Improve performance of Script & OutPoint eq

### LINKED ISSUE #245: Add Dedicated Field-by-Field `eq` and `clone` Methods for Frequently Used Entities
Hey folks at CKB DevRel!!

The current `eq` and `clone` implementations in `Entity.Base` perform full serialization and deserialization via `toBytes()` and `bytesEq()`. For entities that are compared or cloned often (such as `Script` and `Outpoint`), this incurs unnecessary CPU work, memory allocations, and garbage-collection pressure.

## Use Case  
Transaction validation and data transformations frequently compare and clone instances of `Script`, `Outpoint`, and similar entities. A more efficient and transparent approach to equality checks and cloning is needed to improve performance.

## Inefficiency Details  
- **`eq(other)`**  
  1. Serializes `this` into a byte array (O(n)).  
  2. Converts `other` to the same type if needed, then serializes it (O(n)).  
  3. Compares two byte arrays element by element (O(n)).  
  4. Allocates at least two temporary buffers per comparison.  
- **`clone()`**  
  1. Serializes `this` to bytes (O(n)).  
  2. Deserializes those bytes into a new instance (O(n)).  
  3. Allocates intermediate buffers and objects for the new instance.

Combined, these operations scale linearly with the serialized size and generate significant memory churn on each call.

## Proposal  
1. In each frequently used entity class, replace the generic signature with a class-specific one:  
   - In `Script`:  
     ```ts
     eq(other: Script): boolean
     clone(): Script
     ```  
     Compare and copy fields `codeHash`, `hashType`, and `args` directly.  
   - In `Outpoint`:  
     ```ts
     eq(other: Outpoint): boolean
     clone(): Outpoint
     ```  
     Compare and copy fields `txHash` and `index` directly.  

## Expected Benefits
- Runtime Optimization friendly: during garbage collection or JIT compilation, the engine may scan for distinct string objects with identical contents and merge them into one.
- Performance: operations become O(f) where f is the number of fields, with no temporary buffers.
- Clarity: code explicitly shows which fields determine equality and how cloning is performed.

Love & Peace, Phroi %32

phroi:
Closes #245

### DIFF
--- a/packages/core/src/client/client.ts
+++ b/packages/core/src/client/client.ts
@@ -259,54 +259,7 @@ export abstract class Client {
     limit?: NumLike,
     after?: string,
   ): Promise<ClientFindCellsResponse> {
-    const myKey = ClientIndexerSearchKey.from(key);
-    const res = await this.findCellsPagedNoCache(myKey, order, limit, after);
-
-    // If no cells are returned, exit early.
-    if (res.cells.length === 0) {
-      return res;
-    }
-
-    // Assign type and lock scripts based on scriptType.
-    let type = myKey.filter?.script;
-    let lock: typeof type = myKey.script;
-    if (myKey.scriptType === "type") {
-      [lock, type] = [type, lock];
-    }
-
-    // This loop performs string interning on the script fields (codeHash, hashType, args)
-    // for all returned cells. By ensuring that any script component identical to the
-    // source script points to the same string object in memory, we achieve two key benefits:
-    // 1. Memory Reduction: Avoids storing thousands of duplicate strings for common scripts.
-    // 2. Performance Boost: Future equality checks on these fields can become much faster
-    //    (a simple reference check instead of a character-by-character comparison).
-    for (const { cellOutput } of res.cells) {
-      for (const [destination, source] of [
-        [cellOutput.lock, lock],
-        [cellOutput.type, type],
-      ]) {
-        if (!source || !destination) {
-          continue;
-        }
-
-        if (destination.codeHash !== source.codeHash) {
-          continue;
-        }
-        destination.codeHash = source.codeHash;
-
-        if (destination.hashType !== source.hashType) {
-          continue;
-        }
-        destination.hashType = source.hashType;
-
-        if (destination.args !== source.args) {
-          continue;
-        }
-        destination.args = source.args;
-      }
-    }
-
-    // Record cells to the cache and return.
+    const res = await this.findCellsPagedNoCache(key, order, limit, after);
     await this.cache.recordCells(res.cells);
     return res;
   }
--- a/packages/core/src/client/client.ts
+++ b/packages/core/src/client/client.ts
@@ -274,19 +274,35 @@ export abstract class Client {
       [lock, type] = [type, lock];
     }
 
-    // Iterate through the returned cells to intern scripts hex fields from the source.
+    // This loop performs string interning on the script fields (codeHash, hashType, args)
+    // for all returned cells. By ensuring that any script component identical to the
+    // source script points to the same string object in memory, we achieve two key benefits:
+    // 1. Memory Reduction: Avoids storing thousands of duplicate strings for common scripts.
+    // 2. Performance Boost: Future equality checks on these fields can become much faster
+    //    (a simple reference check instead of a character-by-character comparison).
     for (const { cellOutput } of res.cells) {
       for (const [destination, source] of [
-        // Iterate over lock and type
         [cellOutput.lock, lock],
         [cellOutput.type, type],
       ]) {
-        // Intern the hex fields of the Scripts for instant future equality.
-        if (source && destination?.eq(source)) {
-          destination.codeHash = source.codeHash;
-          destination.hashType = source.hashType;
-          destination.args = source.args;
+        if (!source || !destination) {
+          continue;
         }
+
+        if (destination.codeHash !== source.codeHash) {
+          continue;
+        }
+        destination.codeHash = source.codeHash;
+
+        if (destination.hashType !== source.hashType) {
+          continue;
+        }
+        destination.hashType = source.hashType;
+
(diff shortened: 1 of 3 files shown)


packages/core/src/client/client.ts
```diff
@@ -259,7 +259,38 @@ export abstract class Client {
...
+      for (const [destination, source] of [
+        // Iterate over lock and type
+        [cellOutput.lock, lock],
+        [cellOutput.type, type],
+      ]) {
+        // Intern the hex fields of the Scripts for instant future equality.
+        if (source && destination?.eq(source)) {
+          destination.codeHash = source.codeHash;
+          destination.hashType = source.hashType;
+          destination.args = source.args;
```
Hanssen0, line 288: This could cause a problem: the search mode can be in Prefix/Partial, making the result different from our parameters. You can leave the thing as it is, since `hexFrom` already did it for us.

phroi: > This could cause a problem

Not a problem per se, cause we are checking for equality, so either it is equal or not.

We could restrict for **exact only**, but at that point either:
1. We throw an error when it's not equal, which would hamper your code coverage
2. We blindly substitute the script, major error if ever something goes wrong upstream

So what would work and waste the minimal amount of cycles?

**Interleaving check and assignment field by field, this way we dedupe string also in prefix searches.**

I'm gonna go ahead with this and you can review again

Hanssen0: Sorry for the mistake before. You're right, this shouldn't cause a problem because of the equality check.
But why would we need this code? Seems like it does nothing.

If the "dedupe string" you're referring to is using the same string instance for scripts to reduce memory usage, I don't think that's what a `Client` should do. It's hardly possible; `hexFrom` should handle this; The V8 engine might be the place for this. 

This interface only returns ten cells by default each time, and I'm unsure if this will result in a noticeable performance improvement. I would suggest benchmarking to see if this code is necessary.

Hanssen0: image
The V8 engine handles a vast number of strings daily. It's better to let them optimize this rather than we do it manually.

phroi: I'll write the benchmark, not sure how due to the deduping at compilation time, which does not apply here.

I'll think about it

phroi:
I created a `Script.eq` heavy benchmark:

```
import { performance } from "perf_hooks";
import type { Script } from "../ckb/script.js";
import { Cell, type CellLike } from "../ckb/transaction.js";

function createScript(i: number): {
  codeHash: string;
  hashType: string;
  args: string;
} {
  return {
    codeHash: "0x" + (i % 16).toString(16).repeat(64),
    hashType: ("0data" + (i % 3).toString()).replace(/0+/g, ""),
    args: "0x" + ((i + 1) % 16).toString(16).repeat((2 * i) % 170),
  };
}

function createCell(i: number, j: number): Cell {
  return Cell.from(
    JSON.parse(
      JSON.stringify({
        outPoint: {
          txHash: "0x" + (j % 16).toString(16).repeat(64),
          index: "0x1" + (j % 10).toString(16) + "0000000000000000",
        },
        cellOutput: {
          lock: createScript(i),
          type: createScript(i + 1),
          capacity: "0x10000000000000000" + (j % 100).toString(16),
        },
        outputData: "0x1" + (j % 101).toString(16),
      }),
    ) as CellLike,
  );
}

function benchmark(iMax: number, useInterning: boolean): bigint {
  let res = 0n;
  for (let i = 0; i < iMax; i++) {
    const {
      cellOutput: { lock, type },
    } = createCell(i, 0);

    for (let j = 0; j < 100; j++) {
      const cell = createCell(i, j);

      if (useInterning) {
        for (const [destination, source] of [
          [cell.cellOutput.lock, lock],
          [cell.cellOutput.type, type],
        ]) {
          if (!source || !destination) {
            continue;
          }

          if (destination.codeHash !== source.codeHash) {
            continue;
          }
          destination.codeHash = source.codeHash;

          if (destination.hashType !== source.hashType) {
            continue;
          }
          destination.hashType = source.hashType;

          if (destination.args !== source.args) {
            continue;
          }
          destination.args = source.args;
        }
      }

      res = myCalculations(cell, lock, type!, res);
    }
  }

  return res;
}

function myCalculations(
  cell: Cell,
  lock: Script,
  type: Script,
  res: bigint,
): bigint {
  const { cellOutput } = cell;
  const branch = (cellOutput.capacity + res) % 3n;
  if (branch === 0n && cellOutput.lock.eq(lock)) {
    res += 2n;
  } else if (branch === 1n && cellOutput.type?.eq(type)) {
    res += 3n;
  } else if (branch === 2n && cellOutput.capacity % 2n === 0n) {
    res += 5n;
  } else {
    res += 7n;
  }

  if (res % 8n > 0n) {
    return myCalculations(cell, lock, type, res);
  }

  res %= 11n;

  return res;
}

const n = 40000;

let res = benchmark(10, false);
console.log(res);
let start = performance.now();
res += benchmark(n, false);
console.log(performance.now() - start, res);

res = benchmark(10, true);
console.log(res);
start = performance.now();
res += benchmark(n, true);
console.log(performance.now() - start, res);

```

phroi:
Results are the following:
- Field-by-field `Script.eq` comparison yields a 5x speed-up (PR vs `"@ckb-ccc/core": "^1.9.0"`)
- Manual interning of Script fields yields a ~1% slow-down, so I removed it

Love & Peace, Phroi %176

packages/core/src/client/client.ts
phroi: Benchmark at hand, you were right

Hanssen0 (approved):
LGTM

Hanssen0:
Thank you for your contribution and the benchmark results! If you have time, the benchmark scripts are also welcome to submit to the repository so we can easily evaluate future changes.

PR #250, opened 2025-08-12
Second attempt at UDT Manager

### LINKED ISSUE #227: [FR] Merge iCKB Utils UDTHandler features into CCC's Udt

phroi:
Closes #227

### DIFF
--- a/packages/udt/src/udt/index.ts
+++ b/packages/udt/src/udt/index.ts
@@ -105,8 +105,9 @@ export class UdtManager {
   ): Promise<Info> {
     let info = initialInfo;
     for (const cell of cells) {
+      const oldInfo = info;
       info = await this.infoFrom(client, info, cell);
-      if (info.count > 0) {
+      if (info.count > oldInfo.count) {
         tx.addInput(cell);
       }
     }
@@ -126,7 +127,6 @@ export class UdtManager {
     );
     const cellOutput = ccc.CellOutput.from(
       {
-        capacity: amount,
         lock,
         type: this.script,
       },
@@ -153,7 +153,7 @@ export class UdtManager {
     }
 
     const info = await this.infoFrom(client, undefined, cell);
-    if (!info) {
+    if (info.count === 0) {
       throw Error("Change output must be a UDT cell");
     }
 
@@ -267,10 +267,7 @@ export class UdtManager {
       yield { addedInput, initialInput, output };
     }
 
-    throw new ErrorUdtInsufficientCoin(
-      output.balance - addedInput.balance,
-      this,
-    );
+    throw new ErrorUdtInsufficientCoin(minBalance - addedInput.balance, this);
   }
 
   async completeChangeTo(


phroi:
In the end I based it on master, later on we can get back all the other goodies from your PR. OFC the code is not yet ready to be merged, this is more of a proof of concept to gather feedback.

@Hanssen0 could you review and point out what you dislike in this approach?

packages/udt/src/udt/index.ts
```diff
@@ -1,306 +1,397 @@
...
+      count: 0,
+    },
+    ...cells: ccc.Cell[]
+  ): Promise<Info> {
+    let info = initialInfo;
+    for (const cell of cells) {
+      info = await this.infoFrom(client, info, cell);
+      if (info.count > 0) {
+        tx.addInput(cell);
       }
```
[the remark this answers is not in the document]
phroi: True, I'll fix that later

packages/udt/src/udt/index.ts
```diff
@@ -30,277 +1,397 @@
...
-            type: this.script,
-          },
-          ccc.numLeToBytes(amount, 16),
-        );
-      }
-      resTx = ssri.ExecutorResponse.new(transfer);
+    const info = await this.infoFrom(client, undefined, cell);
+    if (!info) {
+      throw Error("Change output must be a UDT cell");
     }
```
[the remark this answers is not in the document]
phroi: True, I'll fix that later

packages/udt/src/udt/index.ts
```diff
@@ -30,277 +1,397 @@
...
+  ): AsyncGenerator<{ addedInput: Info; initialInput: Info; output: Info }> {
+    const initialInput = await this.getInputsInfo(client, tx);
+    let addedInput = await this.infoFrom(client);
+    const output = await this.getOutputsInfo(client, tx);
+
+    const minBalance = output.balance - initialInput.balance;
+    const minCapacity = ccc.numMin(
+      output.capacity - initialInput.capacity,
+      -(await tx.getFee(client)),
+    );
```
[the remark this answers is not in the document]
phroi: While the one calculated on the transaction is the global one

packages/udt/src/udt/index.ts
```diff
@@ -30,277 +1,397 @@
...
+    }
+
+    if (initialInput.count + addedInput.count === 0 && output.count === 0) {
+      return {
+        addedInputs: 0,
+        addedOutputs: 0,
+      };
+    }
+
+    const burned = initialInput.balance + addedInput.balance - output.balance;
```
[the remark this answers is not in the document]
phroi: burned is fine

packages/udt/src/udt/index.ts
```diff
@@ -30,277 +1,397 @@
...
+      }
+    }
+
+    if (addedInput.balance >= minBalance && addedInput.capacity < minCapacity) {
+      yield { addedInput, initialInput, output };
     }
 
-    return tx;
+    throw new ErrorUdtInsufficientCoin(
+      output.balance - addedInput.balance,
```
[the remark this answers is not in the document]
phroi: True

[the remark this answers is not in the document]
phroi: Done

packages/udt/src/udt/index.ts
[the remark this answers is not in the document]
phroi: Done

packages/udt/src/udt/index.ts
```diff
@@ -30,277 +1,397 @@
...
+    },
+    ...cells: ccc.Cell[]
+  ): Promise<Info> {
+    let info = initialInfo;
+    for (const cell of cells) {
+      info = await this.infoFrom(client, info, cell);
+      if (info.count > 0) {
+        tx.addInput(cell);
       }
     }
```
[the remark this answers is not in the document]
phroi: Done

packages/udt/src/udt/index.ts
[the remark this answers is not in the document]
phroi: Done

phroi:
Strange our overlord repeated the same issues two times :man_shrugging:

Hanssen0:
> In the end I based it on master, later on we can get back all the other goodies from your PR. OFC the code is not yet ready to be merged, this is more of a proof of concept to gather feedback.
> 
> @Hanssen0 could you review and point out what you dislike in this approach?

Let me quote your comment on Discord to allow others understand what we're talking about better:

> Mixing SSRI implementation details into a general UDT class is like trying to fit one foot into two shoes.
>
> 1. Solution: Make two classes, Udt and SsriUdt:
> ✅  I'm ok with making most methods in base Udt async so they can sub-classed in SsriUdt, no way around that
> ❌ I'm not ok if base methods signatures contain SSRI details, that's a leaky abstraction
> ❌  I'm not ok if base Udt class cannot express basic metadata of the Udt as properties, just query this metadata in `SsriUdt.from` at creation time
>
> 2. `CompleteX` methods are excessively difficult to understand due to the use of closures in parameters. The ship already sailed for `ccc.Transaction` `CompleteX` methods, but here we can still make it better.
>
> Basic properties of interest relating to UDT cells are:
> * balance (UDT value)
> * capacity
> * count (how many cells are we adding?)
>
> Instead of function closures, we can use a generator and we hard-code these properties as things of interest when adding cells, it simplifies everything. Any code that needs a closure to make decisions based on other computed cell properties, may not be able to fit in the closure approach anyway.
>
> 3. At the beginning there was not enough docs, we agreed on the issue and fixed it. Now we have so much TypeDoc that makes the code difficult to follow the code logic across methods. Let's trim it down a bit.

packages/udt/src/udt/index.ts
```diff
@@ -1,306 +1,394 @@
...
+    public readonly symbol: string,
+    public readonly decimals: number,
+    public readonly icon: string,
+  ) {}
 
-    return ssri.ExecutorResponse.new(undefined);
-  }
+  async *find(
+    client: ccc.Client,
+    locks: ccc.Script[],
```
Hanssen0, line 27: I'll always prefer `Signer`. While it might not be the best name for all cases (Like `Watcher` or sth might be better than `SignerReadonly`), it does reduce the aspects that newcomers have to learn. Besides, the utilities implemented in `Signer` also help us to avoid repeating ourselves.

We should use something like `SignerCkbScriptListReadonly` instead of `Script[]` everywhere, for special cases.

packages/udt/src/udt/index.ts
```diff
@@ -1,306 +1,394 @@
...
-          script: this.script,
-          ...context,
-        },
-      );
-      if (res) {
-        return res.map((res) => ccc.numFromBytes(res));
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async infoFrom(
+    _client: ccc.Client,
+    initialInfo = {
```
Hanssen0, line 65: I personally dislike this design. My principle is that we should always obtain information from the original `Transaction` (the original one defined by CKB, not CCC's) if possible. An example is the `spore` package, where all cobuild things are hidden behind the `Transaction.witnesses`.

Yes, extra information does make implementation simpler, and it will likely improve performance. However, the increased complexity is too painful. It's a trade-off between performance/simplicity, and usability in most cases.

packages/udt/src/udt/index.ts
```diff
@@ -1,306 +1,394 @@
...
-          ccc.mol.Uint128Vec.encode(transfers.map(({ amount }) => amount)),
-        ],
-        {
-          script: this.script,
-        },
-      );
-      if (res) {
-        resTx = res.map((res) => ccc.Transaction.fromBytes(res));
-      }
+  async updateOutput(
```
Hanssen0, line 144: Definitely practical method for updating an output's balance!

packages/udt/src/udt/index.ts
phroi: @Hanssen0 where in the original PR was mentioned that CoBuild is a target?

>  we should always obtain information from the original Transaction

Let's break down what it entails:
- **inputs**: the client is exactly there as initial parameter for obtaining information about the original Transaction, like the inclusion block and **Header**.
- **outputs**: we need to update the parameter signature.

```typescript
    ...cells: {
      cellOutput: ccc.CellOutput;
      outputData: ccc.Hex;
      outPoint?: ccc.OutPoint;
    }[]
```

becomes:

```typescript
    ...cells: ccc.Cell[] | {tx : ccc.Transaction, index: number}[]
```

Not really an issue as you can see.

packages/udt/src/udt/index.ts
phroi: Do you have any particular issue in mind?

Hanssen0: No. I don't think this part has any issues. Just want you to know that I think it's a good idea to have this.

packages/udt/src/udt/index.ts
Hanssen0: We can have them.

packages/udt/src/udt/index.ts
Hanssen0: Let me clarify: the `Info` is an extra thing that devs must maintain. When using the API, invokers must keep the info along with the tx, even if we can get all we need from the tx. It makes things complex.

Hanssen0:
> The issues you mentioned are a minor code change, and we can address them. Any other issue **in the code**?

I'm unsure about what you mean by **in the code**. For me, this currently seems neither completed nor compatible with the original API. I can only provide some suggestions about the basic design.

Do you think we should dive deeper into any particular part?

packages/udt/src/udt/index.ts
phroi: Underlying issue is that CCC doesn't offer a good way to attach additional information to entities like Transaction, Cell and Script... As you know I have been struggling with this since the beginning due to the iCKB headers.

I recently switched to client+cache for the additional information and I was hoping that it would be the end of it.

The CoBuild info would be the object that later on gets serialized into the message?

We can just use a global CCC WeakMap to add this extra info, you just need to modify clone and copy to persist this information across copies

Hanssen0: Besides letting devs focus their mind on a simple tx, this also gives extra compatibility across SDK/ABI/API. Imagine that we're dealing with two devices cooperating to build a transaction. If we have additional information attached to the transaction, we will need to design the ABI carefully to carry it. Alternatively, we need to "pack" this information into the transaction and then "unpack" it on the other end.

Both ways are annoying & complex & leaking abstraction. That's why I insist that it's necessary to avoid attaching information to the transaction.

phroi:
I can see that this PR has zero chances of being merged.

Let's get back to your original PR

PR #260, opened 2025-08-16
perf(core): optimize `Transaction.completeFee`

Hanssen0:


phroi:
I see, you moved `tx.clone` away from the beginning of the loop, so we avoid it whenever possible

packages/core/src/ckb/transaction.ts
```diff
@@ -1998,21 +1998,22 @@ export class Transaction extends mol.Entity.Base<
...
+    // ===
+    // Usually, for the worst situation, three iterations are needed
+    // 1. First attempt to complete the transaction.
+    // 2. Not enough capacity for the change cell.
+    // 3. Fee increased by the change cell.
+    // ===
     while (true) {
-      const tx = this.clone();
-      const collected = await (async () => {
+      collected += await (async () => {
```
phroi, line 2010: Just one question: why not completely dropping the internal anonymous function?

```typescript
  if (options?.shouldAddInputs ?? true) {
    try {
      collected += await this.completeInputsByCapacity(
        from,
        leastFee + leastExtraCapacity,
        filter,
      );
    } catch (err) {
      if (
        err instanceof ErrorTransactionInsufficientCapacity &&
        leastExtraCapacity !== Zero
      ) {
        throw new ErrorTransactionInsufficientCapacity(err.amount, {
          isForChange: true,
        });
      }

      throw err;
    }
  }
```

phroi: This is indeed a realistic scenario

phroi:
It looks good to me!!

PR #262, opened 2025-08-16
feat(core): add `CellAny`

Hanssen0:
It's definitely a mistake to name `CellOnChain` `Cell`, but there is nothing we can do with that right now. To avoid more duplicate code, `CellAny` was added to represent a cell that's on-chain or off-chain.

### DIFF
--- a/.changeset/salty-apples-check.md
+++ b/.changeset/salty-apples-check.md
@@ -1,0 +1,9 @@
+---
+"@ckb-ccc/core": minor
+"@ckb-ccc/ssri": patch
+---
+
+feat(core): add `CellAny`
+
+It's definitely a mistake to name `CellOnChain` `Cell`, but there is nothing we can do with that right now. To avoid more duplicate code, `CellAny` was added to represent a cell that's on-chain or off-chain.
+  
--- a/packages/core/src/ckb/transaction.ts
+++ b/packages/core/src/ckb/transaction.ts
@@ -198,7 +198,7 @@ export class OutPoint extends mol.Entity.Base<OutPointLike, OutPoint>() {
  * @public
  */
 export type CellOutputLike = {
-  capacity: NumLike;
+  capacity?: NumLike | null;
   lock: ScriptLike;
   type?: ScriptLike | null;
 };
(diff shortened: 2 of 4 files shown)


Hanssen0:
Finally, we make our overlord satisfied.
@phroi You know what, we should just assume the not-provided `outputData` is empty. Fighting with type-checking things is killing me 😭.

phroi:
I wonder if there is a better name for CellAny

phroi:
Also it would be nice a method for promotion to ccc.Cell by checking Client

Hanssen0:
> I wonder if there is a better name for `CellAny`

This is the best I can get. Any suggestions are welcome.

> Also it would be nice a method for promotion to ccc.Cell by checking Client

I think we can't get the `outPoint` from `cellOutput` and `outputData`?

phroi:
> I think we can't get the outPoint from cellOutput and outputData?

That's correct, I realized one minute too late :rofl:

phroi:
Just the hash may change as the original transaction is modified

phroi:
Not if we use this CellAny as template to add output cells, ok

phroi:
I stand by my comment: all cells (input and output) in the context of a Transaction are ccc.Cell

Just the hash of outputs may change as the Transaction is updated

phroi:
So we may actually not need ccc.CellAny

Hanssen0:
> So we may actually not need ccc.CellAny

I would say it's still needed. Putting off chain outpoints into `Cell` does make sense, but will bring lots of trouble. Every place using `Cell.outPoint` will have to check if it's valid or not, making things more painful.

Hanssen0:
> > [CellAny] is the best I can get. Any suggestions are welcome.
> 
> Do you fancy any of these? I asked help to GPT-o4 mini 😁
> 
> * CellTransitional
> * CellMixed
> * CellSpectrum
> * CellSuperposition
> * CellBlend

Yea I also ask Gemini... didn't get a better name. Other names like `CellEntity` `CellRecord` `CellObject`... they are longer but do not provide more information than `CellAny` (even less).

phroi:
Ok, the concept is sound, the code looks in order and well thought. 

Just one additional question: would it make sense to add an additional parameter?

Something along the lines of `type : "input"|"output"` and tie to it the value of outpoint?

`{ type : "input", outpoint : ccc.Outpoint} | {type:"output", outpoint : ccc.Outpoint | undefined}`

Hanssen0:
> Ok, the concept is sound, the code looks in order and well thought.
> 
> Just one additional question: would it make sense to add an additional parameter?
> 
> Something along the lines of `type : "input"|"output"` and tie to it the value of outpoint?
> 
> `{ type : "input", outpoint : ccc.Outpoint} | {type:"output", outpoint: ccc.Outpoint | undefined}`

Considering that `CellAny` is still mainly used as a parameter type, I would like to keep it as simple as possible.
Besides, all inputs are outputs before (except cellbases), all outputs may become inputs someday, aren't they?

phroi:
Yeah, this may be more of a concept related to the code consuming these cells like `UDT.balanceFrom/infoFrom`

Looks good to me, feel free to merge!!
Phroi %41

PR #296, opened 2025-08-29
fix(MapLru): use internal Set to track order

### LINKED ISSUE #293: Bug: `ClientCacheMemory.findCells` never stops
Introduced by #244

https://github.com/ckb-devrel/ccc/blob/f52d761dd32270d8943ad664ad2771ebc7ea1229/packages/core/src/client/cache/memory.ts#L103-L117

https://github.com/ckb-devrel/ccc/blob/f52d761dd32270d8943ad664ad2771ebc7ea1229/packages/core/src/client/cache/memory.advanced.ts#L178-L191

We invoke `Map.get` while iterating, which causes the content of `Map` to be rearranged. While the new cells are always added to the end of the Map, this causes the loop to access the same cell repeatedly.

I think it's a problem of the implementation of `MapLru`, modifying itself with a `get` is not expected.

cc @phroi

phroi:
- [ x] I have read the [**Contributing Guidelines**](CONTRIBUTING.md)

Closes #293

### DIFF
--- a/.changeset/forty-feet-worry.md
+++ b/.changeset/forty-feet-worry.md
@@ -1,0 +1,6 @@
+---
+"@ckb-ccc/core": patch
+---
+
+fix(MapLru): use internal Set to track order
+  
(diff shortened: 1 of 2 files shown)


phroi:
Closing in favor of #295

PR #328, opened 2025-11-03
feat: new layer of `FeePayer` as a helper for transaction final completion step

ashuralyk:
# Description
Add a new layer named `FeePayer` (could be changed further), to abstract our transaction complete step, like balancing process, paying with capacity margin or with DeFi swap ratio, etc.

This change can bring much clearer layer hierarchy on CCC, it makes transaction completion step programmable.

### DIFF
--- a/packages/core/src/ckb/transaction.test.ts
+++ b/packages/core/src/ckb/transaction.test.ts
@@ -1268,9 +1268,7 @@ describe("Transaction", () => {
         mockFeePayer2.completeTxFee as ReturnType<typeof vi.fn>
       ).mock.calls[0];
       expect(completeTxFee1Call[0]).toBeInstanceOf(ccc.Transaction);
-      expect(completeTxFee1Call[1]).toBe(client);
       expect(completeTxFee2Call[0]).toBeInstanceOf(ccc.Transaction);
-      expect(completeTxFee2Call[1]).toBe(client);
 
       // Verify prepareTransaction was called before completeTxFee
       // by checking the order of calls
@@ -1437,10 +1435,7 @@ describe("Transaction", () => {
       expect(prepareCallArg).toBeInstanceOf(ccc.Transaction);
       expect(prepareCallArg.outputs.length).toBe(1);
       // completeTxFee should be called with the modified transaction returned by prepareTransaction
-      expect(mockFeePayer1.completeTxFee).toHaveBeenCalledWith(
-        modifiedTx,
-        client,
-      );
+      expect(mockFeePayer1.completeTxFee).toHaveBeenCalledWith(modifiedTx);
     });
   });
 });


packages/core/src/ckb/transaction.ts
```diff
@@ -2410,6 +2295,18 @@ export class Transaction extends mol.Entity.Base<
     return this.completeFeeChangeToLock(from, script, feeRate, filter, options);
   }
 
+  async completeByFeePayer(
+    client: Client,
+    ...feePayers: FeePayer[]
+  ): Promise<void> {
+    for (const feePayer of feePayers) {
+      await feePayer.prepareTransaction(this);
```
Hanssen0: I agree with Gemini here: we should use the return value of `prepareTransaction` rather than assuming it modifies the tx in place. The previous implementation should be considered a bug.

Possible situation: A wallet (like JoyID, a web wallet) decides to handle the `prepareTransaction` stage itself to support more lock scripts in the future, and we serialize the tx for the API. This prevents the operation from happening in place.

To fix this problem without changing APIs, I assume we can use `Transaction.copy` to modify `this` at the end of `completeBy*` methods.

Hanssen0 (changes requested):

packages/core/src/signer/feePayer/feePayer.ts
```diff
@@ -0,0 +1,53 @@
...
+  options?: {
+    feeRateBlockRange?: NumLike;
+    maxFeeRate?: NumLike;
+  };
+}
+
+export abstract class FeePayer {
+  constructor() {}
+
+  abstract completeTxFee(
```
Hanssen0: Do you think it's better to make `completeTxFee` a pure function `(tx: TxLike) => Tx`? Modifying a function argument in place is always confusing and problem-causing.

Also, this helps us to relax the parameter type restriction from `Tx` to `TxLike`.

packages/core/src/signer/feePayer/feePayerGroup.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(private feePayers: FeePayer[]) {
+    super();
+  }
+
+  push(...payers: FeePayer[]): FeePayerGroup {
+    this.feePayers.push(...payers);
+    return this;
+  }
+
+  pop(): FeePayer | undefined {
```
Hanssen0: It's hard to see the use case where we need to modify the fee payers after creating the group. I think we should just remove the `push` & `pop` method and let devs specify them before constructing.

packages/core/src/signer/signer/index.ts
```diff
@@ -78,8 +78,10 @@ export class Signature {
  * This class provides methods to connect, get addresses, and sign transactions.
  * @public
  */
-export abstract class Signer {
-  constructor(protected client_: Client) {}
+export abstract class Signer extends FeePayerFromAddress {
+  constructor(protected client_: Client) {
```
Hanssen0: I do think it should make our life easier if we move this `client_` member to `FeePayer`, so that we can eliminate those `client` parameters appearing in every method.

Hanssen0 (changes requested):
Great design for the `FeePayer` aspect! This should help us reduce future workload. However, this PR still needs to be polished to be developer-friendly.

An important note here is that `TypeLike` is usually for parameter types, while `Type` is for member/variable types. If a member of a new type has a `TypeLike` type, it almost always means the new type should also be a `TypeLike`.

I suggest making new option types here `TypeLike`, and adding corresponding new `Type` for them. For example:

```ts
export interface FeeRateOptionsLike {
  feeRate?: NumLike;
  options?: {
    feeRateBlockRange?: NumLike;
    maxFeeRate?: NumLike;
  };
}

export class FeeRateOptions {
  constructor(
    public readonly feeRate?: Num;
    public readonly options?: {
      feeRateBlockRange?: Num;
      maxFeeRate?: Num;
    }
  ) {}

  ...
}
```

packages/core/src/signer/signer/index.ts
ashuralyk: So, your suggestion is about replacing extending `Signer` with `FeePayerFromAddress` with just injecting an abstracted parameter `FeePayer` to its constructor?

PR #329, opened 2025-11-11
feat(app): signature textarea for signing

Hanssen0:

### DIFF
--- a/packages/demo/src/app/connected/(tools)/Sign/page.tsx
+++ b/packages/demo/src/app/connected/(tools)/Sign/page.tsx
@@ -5,6 +5,7 @@ import { ButtonsPanel } from "@/src/components/ButtonsPanel";
 import { Textarea } from "@/src/components/Textarea";
 import { useApp } from "@/src/context";
 import { ccc } from "@ckb-ccc/connector-react";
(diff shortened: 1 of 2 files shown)


yixyxiu:
LGTM, 👍

Hanssen0:
> After completing the signature, a confirmation dialog appears along with a red error warning. This may worry users, making them think the app is trying to access the clipboard without permission. So instead of automatically copying the signed message to the clipboard, would it be better to provide a clear “Copy” button and let users trigger the action themselves?

Thank you for your testing! This dialog doesn't appear on my side, so I ignored this part. I will modify its behaviour to make it more straightforward that a user is copying the signature.

yixyxiu:
> > After completing the signature, a confirmation dialog appears along with a red error warning. This may worry users, making them think the app is trying to access the clipboard without permission. So instead of automatically copying the signed message to the clipboard, would it be better to provide a clear “Copy” button and let users trigger the action themselves?
> 
> Thank you for your testing! This dialog doesn't appear on my side, so I ignored this part. I will modify its behaviour to make it more straightforward that a user is copying the signature.

Cool, LGTM now.

PR #330, opened 2025-11-18
feat: add superise wallet

rivjc1025:

### DIFF
--- a/packages/superise/src/ckb/index.ts
+++ b/packages/superise/src/ckb/index.ts
@@ -157,7 +157,7 @@ export class CkbSigner extends ccc.Signer {
           scriptInfo.script,
           this.client,
         );
-        if (!index) return;
+        if (typeof index !== "number") return;
 
         indexes.push(index);
       },


packages/superise/src/ckb/index.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  override async signTransaction(
+    tx: ccc.TransactionLike,
+  ): Promise<ccc.Transaction> {
+    const preparedTx = await this.prepareTransaction(tx);
+    return this.signOnlyTransaction(preparedTx);
+  }
+
+  override async sendTransaction(tx: ccc.TransactionLike): Promise<ccc.Hex> {
```
Hanssen0: It seems unnecessary to override `signTransaction` and `sendTransaction`, since they're identical to the default implementations.

Hanssen0 (commented):
Thank you for your contribution! It's happy to see CCC can help devs to integrate the SuperRISE wallet.

This should be a PR that works. However, I have some concerns about the security design of APIs. CCC, or other wallet SDKs, run in an insecure environment, which means code involving security/verification should not depend on the SDK's correctness.

BTW, where can I download the SupeRISE wallet to test this PR?

rivjc1025:
> BTW, where can I download the SupeRISE wallet to test this PR?

The app is still in the testing phase, so we may not be able to provide a version for you to test at the moment. 😂

packages/superise/src/ckb/index.ts
```diff
@@ -0,0 +1,176 @@
...
+import type { Bridge, CkbConnection } from "../advancedBarrel";
+
+export class CkbSigner extends ccc.Signer {
+  get type() {
+    return ccc.SignerType.CKB;
+  }
+
+  get signType() {
+    return ccc.SignerSignType.CkbSecp256k1;
+  }
```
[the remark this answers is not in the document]
fghdotio: Good catch by Gemini here.

packages/superise/src/ckb/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    const tx = ccc.Transaction.from(txLike);
+
+    const scripts = await this.getRelatedScripts(tx);
+
+    await Promise.all(
+      scripts.map(async ({ script, cellDeps }) => {
+        await tx.prepareSighashAllWitness(script, 65, this.client);
+        await tx.addCellDepInfos(this.client, cellDeps);
+      }),
+    );
```
fghdotio: Since `tx` is a shared mutable object, concurrent modifications via `Promise.all` cause race conditions. We should switch to sequential execution to ensure data integrity.

Hanssen0:
Outdated

PR #333, opened 2025-11-30
feat(udt): add udt info querying methods (fix #261 merge conflicts)

phroi:
Hey @Hanssen0, in this PR I fixed #261 minor merge conflicts: feel free to take the code in this branch and force push it on #261 as your own.

I tried to commit on #261, but I don't have the permissions.

Phroi %150

### DIFF
--- a/packages/udt/src/udt/index.ts
+++ b/packages/udt/src/udt/index.ts
@@ -2,6 +2,296 @@ import { ccc } from "@ckb-ccc/core";
 import { ssri } from "@ckb-ccc/ssri";
 
 /**
+ * Error thrown when there are insufficient UDT coins to complete a transaction.
+ * This error provides detailed information about the shortfall, including the
+ * exact amount needed, the UDT type script, and an optional custom reason.
+ *
+ * @public
+ * @category Error
+ * @category UDT
+ *
+ * @example
+ * ```typescript
+ * // This error is typically thrown automatically by UDT methods
+ * try {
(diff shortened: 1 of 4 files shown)


packages/udt/src/udt/index.ts
phroi: @ashuralyk as you can see from here I personally did not make any change to `numLeToBytes` in `Mint`, it just got updated to the current master version, see:

https://github.com/ckb-devrel/ccc/blob/master/packages/udt/src/udt/index.ts#L265

phroi: @gpBlockchain & @Hanssen0 made & merged that change in master with:

https://github.com/ckb-devrel/ccc/commit/b371b07e67f295129defc36190741ab4d783dd96

phroi:
> What I found isn't in any changed code blocks, so I cannot put my comment through Github review tools, I just put them in this comment zone, and to be honest, I'm not sure why those difference exist

Exactly, you couldn't find them because I didn't make them, both changes were made **outside** this commit by commits that happened between the old master (when #261 was created) and the new master (when I started #333), so they just reflect the updated master.

Would you like to start a review on the code itself?

Love & Peace, Phroi %28

phroi:
My target was to resolve #261 merge conflicts and I'm glad that we achieved that. If so @Hanssen0 wishes, she can force push this code to #261

Wish you a good day too :hugs:
Phroi

PR #348, opened 2026-01-09
feat: add register ssri method

ashuralyk:

related contract repo: https://github.com/ashuralyk/ssri-udt


Hanssen0 (commented):
Overall, LGTM. I can see that we add a new UDT implementation called `SSRIUDT` with an SSRI method `SSRIUDT.create`. What is confusing me is about the naming. Instead of the original names, we use `UdtRegister` and `register` here, so I wonder if there is a specified reason for this.

I also checked the script source code, and it looks like there is still some redundancy. Some cleanup might be required.

PR #356, opened 2026-01-27
chore: demo features script deloyer

ashuralyk:

### DIFF
--- a/packages/demo/src/app/connected/(tools)/DeployScript/page.tsx
+++ b/packages/demo/src/app/connected/(tools)/DeployScript/page.tsx
@@ -9,7 +9,7 @@ import { useApp } from "@/src/context";
 import { formatString, useGetExplorerLink } from "@/src/utils";
 import { ccc } from "@ckb-ccc/connector-react";
 import { Loader2, Upload, X } from "lucide-react";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 function formatFileSize(bytes: number): string {
   if (bytes === 0) return "0 Bytes";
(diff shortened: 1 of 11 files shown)


packages/demo/src/app/connected/(tools)/DeployScript/page.tsx
```diff
@@ -0,0 +1,750 @@
...
+  const typeIdArgs = cell.cellOutput.type?.args || "";
+  const dataSize = cell.outputData ? ccc.bytesFrom(cell.outputData).length : 0;
+
+  return (
+    <BigButton
+      key={ccc.hexFrom(cell.outPoint.toBytes())}
+      size="sm"
+      iconName="FileCode"
+      onClick={onSelect}
+      className={isSelected ? "border-purple-500 bg-purple-50" : ""}
```
Hanssen0: The bg-purple-500 seems not working.

packages/demo/src/app/connected/(tools)/DeployScript/page.tsx
```diff
@@ -0,0 +1,750 @@
...
+      } catch (err) {
+        console.error("Error scanning Type ID cells:", err);
+      } finally {
+        setIsScanningCells(false);
+      }
+    })();
+  }, [signer]);
+
+  // Compare addresses when both are available
+  useEffect(() => {
```
Hanssen0: It's better to use `useMemo` here.

Hanssen0 (commented):
This new tool overall looks good to me. The problem is that the code seems overcomplicated (640-line function). While it brings a fancy visual effect, I suggest refactoring this code to make it less coupled.

packages/demo/src/components/Message.tsx
```diff
@@ -66,9 +66,9 @@ export function Message({
               }
         }
       >
-        <p className={`text-sm break-all whitespace-pre-wrap ${colorClass}`}>
+        <div className={`text-sm break-all whitespace-pre-wrap ${colorClass}`}>
```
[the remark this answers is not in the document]
ashuralyk: Because embedding `div` under a tag `p` would trigger warning prompt in Application, so I made this change for better compatibility, and indeed it doesn't bring other problems to other components.

Hanssen0:
See #466

PR #360, opened 2026-03-02
feat(core): extract SignerMultisigCkbBase for Omnilock multisig

phroi:
## Why

The CKB multisig signer and Omnilock multisig signer share ~46% of their code. The only real differences are witness encoding (plain multisig bytes vs. OmniLockWitnessLock envelope) and script cardinality (multiple variants vs. one).

## Changes

- Extract `MultisigCkbWitness` entity into standalone module
- Create abstract `SignerMultisigCkbBase` with `encodeWitnessLock`/`decodeWitnessLock` template methods
- Slim `SignerMultisigCkbReadonly` to thin subclass
- Add `SignerMultisigOmniLockReadonly` and `SignerMultisigOmniLockPrivateKey` for Omnilock auth flag 0x06
- Add `OmniLockWitnessLock` molecule encoding utility

Builds on #349.


phroi:
After completing the [guard-signing comparison](https://github.com/sonami-tech/rosen-bridge-ckb-integration/blob/master/guard-signing/comparison.md), we are leaning towards ECDSA TSS over Omnilock multisig for the Rosen Bridge CKB integration.

The main reason is that Rosen Bridge already runs ECDSA TSS for Bitcoin, Ethereum, Doge, Binance, and others: same tss-lib, same Go binary, same keygen shares. CKB's default secp256k1-blake160 lock accepts standard ECDSA signatures, so CKB can plug into that existing infrastructure instead of a dedicated signing package and the Omnilock dependency chain (molecule encoding, CCC multisig PR, witness assembly).

Guard rotation (replacing the set of ~30 bridge operators that jointly control custody) is where it matters most: every non-Ergo chain must migrate all assets to a new address when the guard set changes. With TSS, CKB piggybacks on the bridge-wide migration event (all ECDSA chains rotate together using the same tooling and procedures). With Omnilock multisig, CKB would be the only non-Ergo chain migrating through a non-TSS path, requiring its own migration procedures and tooling that no other chain uses.

Closing #360 accordingly. The multisig signers in this PR are still useful for other Omnilock users, just no longer a dependency for us.

Phroi

Hanssen0:
> Closing #360 accordingly. The multisig signers in this PR are still useful for other Omnilock users, just no longer a dependency for us.
> 
> Phroi

Thank you for your contribution! Though it's closed due to the Rosen Bridge's choice, this PR is still meaningful for the ecosystem for adding OmniLock's multisig support. Considering that the new commit doesn't modify APIs, I assume that we can merge the original multisig PR first and add this code later.

Before doing that, I want to ask for your approval: If we want to add the Omnilock's multisig in the future, can we reuse the code in this PR?

phroi:
> Before doing that, I want to ask for your approval: If we want to add the Omnilock's multisig in the future, can we reuse the code in this PR?

OFC, it would be my pleasure!

Phroi

PR #362, opened 2026-03-21
feat(core): `FeePayer` abstraction

Hanssen0:

Refactors #328

### DIFF
--- a/packages/core/src/ckb/transaction.ts
+++ b/packages/core/src/ckb/transaction.ts
@@ -1,13 +1,20 @@
 import { Bytes, BytesLike, bytesFrom } from "../bytes/index.js";
-import type { ClientCollectableSearchKeyFilterLike } from "../client/clientTypes.advanced.js";
 import {
   ClientBlockHeader,
+  ClientCollectableSearchKeyFilterLike,
   type CellDepInfoLike,
   type Client,
   type ClientBlockHeaderLike,
 } from "../client/index.js";
 import { KnownScript } from "../client/knownScript.js";
 import { Codec, Entity, codec } from "../codec/index.js";
+import {
+  FeePayer,
+  FeePayerCompleteFeeChangeFn,
+  FeePayerCompleteInputsContext,
+  FeePayerCompleteInputsOptionsLike,
+  FeePayerFromAddress,
+} from "../feePayer/index.js";
 import { Zero, fixedPointFrom } from "../fixedPoint/index.js";
 import { Hasher, HasherCkb, hashCkb } from "../hasher/index.js";
 import { Hex, HexLike, hexFrom } from "../hex/index.js";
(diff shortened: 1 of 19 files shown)


packages/core/src/feePayer/feePayer.ts
```diff
@@ -0,0 +1,322 @@
...
+      const fee = await tx.getFee(this.client);
+      if (fee < leastFee + leastExtraCapacity) {
+        // Not enough capacity are collected previously
+        throw new ErrorFeePayerInsufficientCapacity(
+          leastFee + leastExtraCapacity - fee,
+          { isForChange: leastExtraCapacity !== Zero },
+        );
+      }
+
+      await this.prepareTransaction(tx);
```
ashuralyk, line 176: since the function `prepareTransaction` has added a return value to update the imported transaction object semantically, I recommend here to align with this principle and alter like this:

```
tx = await this.prepareTransaction(tx);
```

and so as below at line `211`

packages/core/src/feePayer/feePayer.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    // ===
+    // Usually, for the worst situation, three iterations are needed
+    // 1. First attempt to complete the transaction.
+    // 2. Not enough capacity for the change cell.
+    // 3. Fee increased by the change cell.
+    // ===
+    while (true) {
+      try {
+        const res = await this.collectCapacity(
```
ashuralyk, line 146: look at the comment on `collectCapacity`, there's a contract inside remark section to prohibit it to throw errors, but the code here encloses it in a `try...catch` statement and assumes it probably throw errors, although it won't cause any issues, it's obviously a contradiction anyway

packages/core/src/ckb/transaction.ts
```diff
@@ -2157,8 +2127,11 @@ export class Transaction extends Entity.Base<TransactionLike, Transaction>() {
    * ```
    */
   async completeFee(
-    from: Signer,
-    change: (tx: Transaction, capacity: Num) => Promise<NumLike> | NumLike,
+    from: FeePayer<
+      FeePayerCompleteInputsOptionsLike,
+      FeePayerCompleteInputsContext
+    >,
```
ashuralyk, line 2133: is there possible to alias the abstract `FeePayer` with concrete generic parameters to an easy-to-recognize name, I guess it'll be helpful for the readability and reuseablility.

packages/core/src/feePayer/feePayer.test.ts
```diff
@@ -0,0 +1,586 @@
...
+  });
+
+  describe("completeInputs", () => {
+    it("should collect cells using the accumulator", async () => {
+      const tx = ccc.Transaction.from({
+        outputs: [],
+      });
+
+      // Collect 3 cells
+      const { addedCount, accumulated } = await feePayer.completeInputs(
```
ashuralyk, line 86: the function `completeInputs` itself will return a new transaction object but the below assertion uses the above old one, so I guess this sort of assertions should be considered to fix altogether.

ashuralyk (commented):
I highly value the usage of generic parameters in the abstract layer of `FeePayer`, it makes this module more sophisticated than ordinary class hierarchy.

Overall, changes are looking good to me.

PR #466, opened 2026-08-03
feat: demo features script deloyer

Hanssen0:
Recreated for #356 


Hanssen0:
I know, I know, this PR is not perfect currently. But it won't be published to npm as a part of the SDK so we can update it anytime, and I do think the script deployer is useful. I'll just accept it's imperfect.

PR #488, opened 2026-08-18
fix(core): close idle WebSocket transports after requests

### LINKED ISSUE #444: [BUG] Public Client WebSocket Prevents Node Exit
Hello folks at CKB DevRel 🤗 

I noticed that a Node process using CCC's default public client stays alive after its RPC completes. The client opens a WebSocket but provides no way to close it.

## Minimal Reproduction

```js
import { ccc } from "@ckb-ccc/core";

const client = new ccc.ClientPublicTestnet();
await client.getTipHeader();
console.log("done");
```

And to run it:
```sh
timeout 600 node repro.mjs
```

## Behavior

The RPC completes and `done` prints, but the process remains alive until killed:

```text
done
exit=124 elapsed=600.02
```

In Node, the default public client selects WebSocket. A successful response clears its request timer and pending entry but leaves the socket open and referenced. Neither the transport nor client exposes `close()`/`dispose()`, and there is no idle close.

## Environment

- `@ckb-ccc/core@1.17.0`
- Node.js 24.18.0
- Linux

## Impact

A short-lived command can finish its RPC work and print its final output, but still sit until an outer timeout kills it. In CI, cron, or one-shot containers this can turn successful work into a timeout, delay the next run, or cause a supervisor to retry work that already completed.

The problem is more serious when an application creates clients repeatedly. Every used client can leave another WebSocket and file descriptor open. For example, creating one client per second would leave roughly 600 sockets after the ten-minute period reproduced above. Over time this could hit file-descriptor, memory, RPC connection, or upstream endpoint limits. Worker processes multiply the same effect.

Keep up the Great Work 💪 
Phroi %48

qiaoshanglong:
Closes #444

## Summary

- Close a WebSocket transport once its last pending JSON-RPC request completes.
- Keep the connection open while concurrent requests are still pending.
- Add a regression test covering concurrent requests and idle cleanup.
- Include a patch changeset for @ckb-ccc/core.

## Testing

- `pnpm exec vitest run`
- `pnpm exec tsdown`
- targeted ESLint
- targeted Prettier check

### DIFF
--- a/.changeset/quiet-websockets-exit.md
+++ b/.changeset/quiet-websockets-exit.md
@@ -1,0 +1,5 @@
+---
+"@ckb-ccc/core": patch
+---
+
+Close idle WebSocket transports after all pending JSON-RPC requests complete.
--- a/packages/core/src/jsonRpc/transports/webSocket.test.ts
+++ b/packages/core/src/jsonRpc/transports/webSocket.test.ts
@@ -1,0 +1,81 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { FakeWebSocket } = vi.hoisted(() => {
+  class FakeWebSocket {
+    static readonly CONNECTING = 0;
+    static readonly OPEN = 1;
+    static readonly CLOSING = 2;
+    static readonly CLOSED = 3;
+    static readonly instances: FakeWebSocket[] = [];
+
+    readonly close = vi.fn(() => {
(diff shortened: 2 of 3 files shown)


Hanssen0:
Hi @qiaoshanglong! Thank you for making a contribution to the CCC.

The approach you proposed in this PR is not how we should use WebSocket, though. If the use case is closing the connection right after a (batch of) request, Http request should be preferred.

Related to #487

Issue #531, opened 2026-09-12
JoyID mainnet cellDeps are all spent again (recurrence of #288); #277's fix still isn't in


LusoCryptoLabs:
All five mainnet JoyID `cellDeps` point at outputs of `0x8a605a4402cadda69fa64fd25cbbd74058e3eb86a7a72aee3d25df278564d31b`, and all five of those outputs are spent. A mainnet transaction that needs the JoyID lock's cell deps therefore names cells that no longer exist, and is rejected before any script runs.

Verified with `getCellLive` against mainnet on 2026-09-12, in **1.12.5 and in 1.20.1**:

```
@ckb-ccc/core 1.20.1, mainnet JoyID
  codeHash 0xd00c84f0ec8fd441c38bc3f87a371f547190f2fcff88e642bc5bf54b9e318323  hashType type
  SPENT  0x8a605a4402cadda69fa64fd25cbbd74058e3eb86a7a72aee3d25df278564d31b:0
  SPENT  0x8a605a4402cadda69fa64fd25cbbd74058e3eb86a7a72aee3d25df278564d31b:1
  SPENT  0x8a605a4402cadda69fa64fd25cbbd74058e3eb86a7a72aee3d25df278564d31b:2
  SPENT  0x8a605a4402cadda69fa64fd25cbbd74058e3eb86a7a72aee3d25df278564d31b:3
  SPENT  0x8a605a4402cadda69fa64fd25cbbd74058e3eb86a7a72aee3d25df278564d31b:4

  0 live, 5 spent
```

**The code did not change, the cells did.** Following each dep's own type id finds live cells carrying the same data at `0x673e1843035ce9c249efc36b62865e7a7c9f6e2ca6fc9c66fd108307e81355f9:0-4`. Combining the five data hashes (sorted by type id, so dep ordering cannot affect the result) gives the same value before and after the move, which is how we noticed: we fingerprint JoyID's code cells to detect an actual upgrade, and this move did not register as one.

Testnet is unaffected: all five deps there are live and nothing has moved.

### This is a recurrence

- #288 (2025-08-28) is the same failure one JoyID generation earlier, fixed by #287 by updating the outpoints by hand.
- #279 had already split JoyID from a `depGroup` into five `code` deps "to allow for an easy transition during JoyID upgrade".
[...]

PR #540, opened 2026-09-13
docs(core): say that a dep with a type id ignores its configured outpoint

LusoCryptoLabs:
As promised in #531. Documentation only, no behaviour change.

The situation this is meant to prevent: several outpoints in the known-script tables no longer resolve on chain, because a code cell can be consumed and recreated carrying the same code. Anyone who checks them against a node sees spent cells and concludes the library is broken. I did exactly that, and reported it as a bug.

It is not a bug, because nothing uses those outpoints as written: any dep carrying a `type` is resolved live by `getCellDeps` before it reaches a transaction. `getCellDeps` already said so, but in the method, which is not where somebody who is looking at the outpoints is reading.

So this adds:

- **A header on both known-script tables** (mainnet and testnet) saying the outpoint is a starting point rather than a promise of liveness, that a dep with a `type` is resolved before use, and that finding these spent is expected. It also states the maintenance rule that falls out of it: a dep with a type id needs no attention when its code cell moves, and a dep without one does, which is the reason #279 split the JoyID dep group into five.
- **A stronger line in `getCellDeps`**, saying plainly that the configured outpoint is discarded when a type is present, and that the case needing maintenance is a dep without one.

Happy to reword any of it, cut it down, or drop the testnet copy if one is enough. A changeset is included as a patch; say the word if a docs-only change should not carry one and I will remove it.


Hanssen0 (changes requested):
These documentation updates look entirely correct, but I am certain no human would want to read them in their entirety. Please be more concise. The documentation should only cover what matters to developers, rather than simply translating the code. Furthermore, such verbose docstrings increase the likelihood of errors during future code and comment maintenance.

LusoCryptoLabs:
@Hanssen0 Agreed, it was translating the code. Cut to what a developer needs: the docstring says resolved on-chain when found, configured outpoint otherwise, and always as written without a type; the table headers say the same in four lines and point at #531. 88 added lines became 32.

yixyxiu (commented):
LGTM. Thanks for documenting this — the confusion around a "spent" outpoint in the known-script tables not being a bug is worth capturing.

Issue #517, opened 2026-09-09
SignerBtc.signPsbt options (autoFinalized, inputsToSign) are not normalized consistently across wallets, causing divergent behavior


yixyxiu:
### Description

`SignPsbtOptionsLike` (`packages/core/src/signer/btc/psbt.ts`) declares:
- `autoFinalized?: boolean` — documented as "Default is true."
- `inputsToSign?: InputToSignLike[]`

However, the BTC wallet signer implementations handle this inconsistently, so a dApp calling `signer.signPsbt(psbtHex)` (or with an incomplete `options`) can get different results depending on which wallet is connected — even though the caller wrote identical code.

### Where the inconsistency comes from

- **JoyID** (`packages/joy-id/src/btc/index.ts`) correctly normalizes options via `ccc.SignPsbtOptions.from(options)` before use, so `autoFinalized` reliably defaults to `true` as documented.
- **UniSat** (`packages/uni-sat/src/signer.ts`) and **OKX** (`packages/okx/src/btc/index.ts`) pass the raw `options` argument straight through to the injected wallet provider's `signPsbt(psbtHex, options)`:
  ```ts
  async signPsbt(psbtHex: ccc.HexLike, options?: ccc.SignPsbtOptionsLike): Promise<ccc.Hex> {
    return ccc.hexFrom(
      await this.provider.signPsbt(ccc.hexFrom(psbtHex).slice(2), options),
    );
  }
  ```
  When `options` (or `options.autoFinalized`) is `undefined`, the actual default is whatever UniSat's/OKX's own extension implements — not guaranteed to match CCC's documented default of `true`.

  Worse, UniSat's and OKX's real APIs expect the field name `toSignInputs`, not `inputsToSign` (see [UniSat docs](https://docs.unisat.io/developer-support/open-api-documentation/unisat-wallet#signpsbt), [OKX docs](https://okx.github.io/js-wallet-sdk/#/./coins/bitcoin/pbst?id=sign-psbt-with-key-path-and-script-path)). Since `options` is forwarded unmodified, a caller who sets `options.inputsToSign` on these two wallets has it silently ignored — the wallet instead falls back to signing *all* inputs matching the connected address.

- **Xverse** (`packages/xverse/src/signer.ts`) never sends any `autoFinalized`/finalize-equivalent field in its RPC request at all:
  ```ts
  this.provider.request("signPsbt", {
    psbt: psbtBase64,
    signInputs,
    broadcast: false,
  })
  ```
[...]

PR #541, opened 2026-09-16
Fix/normalize sign psbt options

fghdotio:
Addresses #517.

Normalize `signPsbt` options through `ccc.SignPsbtOptions.from()` in every `SignerBtc` and translate them to each wallet's request shape.

JoyID honours `autoFinalized`, but when it is `true` it goes one step further than the other wallets: after finalizing it extracts the transaction and returns the raw transaction hex `0x02000000…` instead of a finalized PSBT. With `autoFinalized: false` it returns a proper PSBT with `partialSig`. This means the default `signPsbt(hex)` result on JoyID cannot be fed back into a PSBT parser, unlike UniSat / OKX / Xverse. That is a return-value contract question rather than option normalization, so it is left for a separate issue.


Hanssen0 (approved):
LGTM

